PR #30, opened 2020-05-24
feat(database): use regex operator to look for directory

eteplus:
Support for query subdirectories

```bash
content/
  articles/
    article-1.md
    article-2.md
    2020/05/article-3.md
  home.md
```

### DIFF
--- a/example/content/articles/2020/05/markdown-example.md
+++ b/example/content/articles/2020/05/markdown-example.md
@@ -1,0 +1,174 @@
+---
+title: Markdown example
+slug: markdown-example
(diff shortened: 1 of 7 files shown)


benjamincanac:
Hey @eteplus,

Thanks for the PR, indeed it could be nice to deep fetch a directory, but maybe it should be optional because if you're fetching the root directory for example, you might not want to get all the files.

It could be an option like `deep`:

```js
await $content('articles', { deep: true }).fetch()
```

What do you think?

eteplus:
@benjamincanac Make sense, It would be more flexible.

PR #65, opened 2020-05-29
Rehype plugins

benjamincanac:


juliomrqz:
@benjamincanac, is it possible to use local plugins in a similar way that Nuxt.js modules? For example:

```js
/// nuxt.config.js
export default {
  content: {
    markdown: {
      remarkPlugins: [
        "~/plugins/remark-plugin.js",
      ],
      rehypePlugins: [
        ["~/plugins/rehype-plugin.js", { option: 2 }],
      ],
    },
  },
};

```

benjamincanac:
@juliomrqz Indeed it will be! You can already try this PR.

docs/components/bases/CodeGroup.vue
```diff
@@ -31,7 +29,7 @@ export default {
     }
   },
   created () {
-    this.$slots.default.map((slot) => {
+    this.$slots.default.filter(slot => !!slot.componentOptions).map((slot) => {
```
Atinux, line 32: Better to use `Boolean(slot.componentOptions)`

docs/components/bases/CodeGroup.vue
```diff
@@ -40,7 +38,7 @@ export default {
     this.activeTabIndex = 0
   },
   mounted () {
-    this.tabs = this.$slots.default.map((slot) => {
+    this.tabs = this.$slots.default.filter(slot => !!slot.componentOptions).map((slot) => {
```
Atinux, line 41: Same here for Boolean

docs/content/en/writing.md
```diff
@@ -92,7 +92,9 @@ It will be transformed to it's JSON AST structure, and by using the `nuxt-conten
 
 ### Links
 
-Links are transformed to add valid `target` and `rel` attributes. You can change this behaviour, see [configuration](/configuration#markdownexternallinks). Relative links are also automatically transformed to `nuxt-link` to provide navigations between page components and enhance performances with smart prefetching.
+Links are transformed to add valid `target` and `rel` attributes using [remark-external-links](https://github.com/remarkjs/remark-external-links). You can check [here](/configuration#markdown) to learn how to configure this plugin.
+
+Relative links are also automatically transformed to `nuxt-link` to provide navigations between page components and enhance performances with smart prefetching.
```
Atinux, line 97: Suggesting to use `[<nuxt-link>](https://nuxtjs.org/api/components-nuxt-link/)`

lib/database.js
```diff
@@ -21,7 +21,7 @@ class Database extends Hookable {
     super()
     this.dir = options.dir || process.cwd()
     this.cwd = options.cwd || process.cwd()
-    this.markdown = new Markdown(options.markdown)
+    this.markdown = new Markdown({ nuxt: options.nuxt, ...options.markdown })
```
Atinux, line 24: That would be a breaking change for using it outside of Nuxt (I already do). new Markdown should not depend on `nuxt` instance.

lib/index.js
```diff
@@ -54,6 +26,7 @@ module.exports = async function () {
   })
 
   const database = new Database({
+    nuxt: this.nuxt,
```
Atinux, line 29: Same here, Database should not depend of `nuxt` instance

lib/parsers/markdown/index.js
```diff
@@ -15,7 +11,43 @@ const jsonCompiler = require('./compilers/json')
...
+
+      if (typeof plugin === 'string') {
+        name = plugin
+        options = this.options[camelCase(name)]
+      } else if (Array.isArray(plugin)) {
+        [name, options] = plugin
+      }
+
+      try {
+        const instance = require(this.options.nuxt.resolver.resolvePath(name))
```
Atinux, line 34: Having `this.resolvePath` could avoid depending of `this.options.nuxt` (default if just returning the name/path)

Atinux: The more I think, the more I believe this sanitization should be outside of the Markdown class but directly where we set the options/defaults.

So new Mardkown() deals with correct plugins syntax: [{ instance, options }, ...]

juliomrqz:
@benjamincanac Perfect, I’ll check it out. Thanks 😎

docs/components/bases/CodeGroup.vue
benjamincanac: 👍

docs/components/bases/CodeGroup.vue
benjamincanac: 👍

docs/content/en/writing.md
benjamincanac: I've used [nuxt-link] instead of [<nuxt-link>]

lib/database.js
benjamincanac: 👍

lib/index.js
benjamincanac: 👍

lib/parsers/markdown/index.js
benjamincanac: 👍

PR #70, opened 2020-05-30
Translated docs into Russian

cogor:


Atinux:
Thank you @cogor 

Could you also update the nuxt.config.js as well as the i18n/ directory to add russian?

cogor:
> Thank you @cogor
> 
> Could you also update the nuxt.config.js as well as the i18n/ directory to add russian?

Yes, of course

cogor:
This file was in the /en directory, index.md and index copy.md are the same

PR #78, opened 2020-06-01
Handle router base

benjamincanac:


lib/templates/plugin.client.js
```diff
@@ -1,6 +1,7 @@
 import Vue from 'vue'
 import NuxtContent from './nuxt-content'
 import QueryBuilder from './query-builder'
+const logger = require('consola').withScope('@nuxt/content')
```
Atinux, line 4: Could you also wrap with `<% if (options.watch) { %>...<% } %>` to avoid importing consola on production?

benjamincanac: Done 👍

Issue #82, opened 2020-06-02
Document how to display content outside component (e.g., RSS)


philipithomas:
Users will want to display their content outside of components, such as in an RSS feed. Currently, the [displaying content](https://content.nuxtjs.org/displaying#component) page only discusses using a component to render content. The document object does not include HTML or a plaintext summary that could be used for this. 

### Describe the solution you'd like

On the `Displaying content` page, document how to include content programmatically, such as in an RSS feed.

rclement:
I solved this very use-case last week when migrating to `@nuxt/content` in my website, along with `@nuxtjs/feed-module` : https://github.com/rclement/romain-clement.net/blob/master/nuxt.config.ts#L28

Almost everything needed is already in the documentation: https://content.nuxtjs.org/advanced#programmatic-usage

A nice addition to the documentation would be a basic code snippet to integrate with `@nuxtjs/feed-module`, which takes care of everything feeds-related automagically.

@philipithomas did you want to start a PR for this or may I start one?

philipithomas:
@rclement Looks like you have a great solution. If you want to open a PR, that would be great! 

I'm busy this week, but if this issue is still open by the weekend, I may try to update the docs using your example code.

philipithomas:
@rclement It looks like your RSS feed returns the `summary` as the content, rather than the body of your markdown. Is that right? 

I'd like to return the entire article body in the `content` section of the RSS.

rclement:
@philipithomas Yes that's right, this is common practice so that potential readers are still coming back to your site (so you can obtain some more meaningful analytics metrics if you're into this stuff). And using `@nuxt/content` you do not have access to a textual version of the body directly, only the JSON AST.

However you have access to the `text` field containing the raw body during the hook `file:beforeInsert` according to: https://content.nuxtjs.org/advanced#contentfilebeforeinsert
You could copy it and/or process it in some way to extract a plain-text body to be injected afterwards in feeds `content` section.

Atinux:
Thank you for your answers @rclement, a PR for integrating with feed module is more thank welcome 🙏

rclement:
@philipithomas @Atinux I started PR #93, feel free to comment and correct it!

benjamincanac:
@rclement Thanks for the PR!

lukeocodes:
Here's me building integration with `nuxtjs/feed` and you merge a PR showing how to do it :D 

[...]

PR #93, opened 2020-06-03
Add feed-module integration in documentation

rclement:
As discussed in issue #82 with @philipithomas and @Atinux, this PR adds an example of integration between `@nuxt/content` and the Nuxt.js feed-module `@nuxtjs/feed`.


Atinux:
Awesome, thank you very much @rclement

PR #113, opened 2020-06-06
chore: remove unused navbar in starters/slides

dtmzr:
While trying to figure out more about this module I found this `navbar.vue` component which seems not to be used anywhere.

So therefore we could remove it? :)


mathieunicolas:
Well it's a component, so it's in the `components` folder, and you can use it or not use it !
When adding the component to the `default` layout, I have a navbar in the layout =)

Atinux:
Indeed, we plan to anyway move these starters as soon as we implement the themes in Nuxt :)

Thank you for the PR and by looking at the source code anyway :)

PR #125, opened 2020-06-12
feat: live editing

benjamincanac:


lib/middleware.js
```diff
@@ -32,8 +33,22 @@ module.exports = ({ ws, database }) => async (req, res) => {
   } else if (req.method === 'GET') {
     params = nodeReq.get(req)
   }
+
+  const filePath = join(dir, url)
+  const stats = await fs.lstat(filePath).catch(_ => null)
+  if (req.method === 'PUT' && stats && stats.isFile()) {
```
Atinux, line 39: We need to check if we are in dev ⚠️

lib/middleware.js
```diff
@@ -32,8 +33,22 @@ module.exports = ({ ws, database }) => async (req, res) => {
...
+    await fs.writeFile(filePath, params.file, 'utf-8')
+
+    return nodeRes.send(req, res, {})
+  }
+
+  if (req.method === 'GET' && stats && stats.isFile()) {
```
Atinux, line 45: Same here I would say

PR #143, opened 2020-06-15
fix: attributes case

benjamincanac:


lib/templates/nuxt-content.js
```diff
@@ -1,3 +1,4 @@
+const { paramCase } = require('change-case')
```
Atinux, line 1: Do we need to add a dependency here?

wooorm:
Btw: [`property-information`](https://github.com/wooorm/property-information) has all the info needed to turn properties in HTML attribute case!

Atinux:
Will look into it, thank you for pointing that out @wooorm

PR #164, opened 2020-06-19
feat: global components

benjamincanac:


docs/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "nuxt-content-docs",
+  "name": "@nuxt/content-docs",
```
Atinux, line 2: Is it necessary?

benjamincanac: Not really actually

Atinux:
```suggestion
  "name": "nuxt-content-docs",
```

docs/components/global/bases/Dropdown.vue
```diff
@@ -19,7 +19,7 @@
 </template>
 
 <script>
-import ClickAway from '../directives/click-away'
+import ClickAway from '~/components/directives/click-away'
```
Atinux, line 22: I suggest moving directives outside of `components`

PR #167, opened 2020-06-20
feat: add .yml support for YAML

azrikahar:
The `.yml` extension for YAML is pretty popular from what I've seen. Perhaps it's partially because it's pretty often used in DevOps tools? 

Tried to search around for a consensus regarding this, but seems like there are avid supporters on both sides. The [YAML official site](https://yaml.org/faq.html) does mention to use `.yaml` when possible as pointed out when I was googling about it, but FWIW [Jekyll Data Files](https://jekyllrb.com/docs/datafiles/) that does similar thing with nuxt/content also supports both `.yml` and `.yaml` extensions so I figured no harm in adding this. 

Tested this in the example project as seen in the commit and it works.

### DIFF
--- a/example/content/yml.yml
+++ b/example/content/yml.yml
@@ -1,0 +1,2 @@
+title: Yml
+description: This is a .yml file
--- a/example/pages/index.vue
+++ b/example/pages/index.vue
@@ -14,6 +14,8 @@
     <pre>{{ csv }}</pre>
     <h2>YAML</h2>
     <pre>{{ yaml }}</pre>
+    <h2>YML</h2>
+    <pre>{{ yml }}</pre>
   </div>
 </template>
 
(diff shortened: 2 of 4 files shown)


benjamincanac:
Hey @azrikahar,

Thanks for your PR! Indeed there is no harm supporting `.yml` files :) 

Could you add a `yml` file in `test/fixture`, like you did in the example so the coverage doesn't decrease?

azrikahar:
@benjamincanac I wasn't really sure how it works (tests & codecov unfortunately still are foreign to me), but I did as you mentioned. I added it in the `test/fixture/content/formats` directory and seems like it does the trick for the coverage. Thank you very much for guiding me :)

benjamincanac:
Hey @azrikahar, yes perfect! Thanks

PR #180, opened 2020-06-25
fix: handle spa

benjamincanac:


Atinux:
Actually they simply need to set `target: 'static'` is they are using `mode: 'spa'`

PR #258, opened 2020-07-13
feat: add classes for live editing

alvarosabu:
## Types of changes

## Description

- Adds `author` to `nuxt-content-container` when is being edited.
- Adds `nuxt-content-editor` class to text area to avoid using `!important`  if you want to re-style the live edit input.

## Checklist:

I think it doesn't need to be pointed out in the documentation, but If you think it does, I'm happy to add it.

### DIFF
--- a/lib/templates/nuxt-content.dev.vue
+++ b/lib/templates/nuxt-content.dev.vue
@@ -92,6 +92,7 @@ export default {
 .nuxt-content-container {
   position: relative;
 }
+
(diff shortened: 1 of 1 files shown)


Atinux:
Thank you for the PR, I would actually rename `author` to `is-editing` instead

alvarosaburido:
Sure thing.

benjamincanac:
Thanks @alvarosaburido!

Issue #95, opened 2020-06-03
How to handle search properly - different result observed


rajesh-h:
Reproduction link - https://codesandbox.io/s/github/rajesh-h/cs_content

quick reproduction:
Click on search icon -> enter text -> Pressure -> Click enter/search
    You will get result with only one article

https://sj1bl.sse.codesandbox.io/search/pressure  - you will get more than one result

Details:
What happens when you click on search?
Search input is present in [header component](https://github.com/rajesh-h/cs_content/blob/ee047a6d60b9f4d41c8a243f23e8b7f6a1e974e8/components/TheHeader.vue#L11)
It invokes a method which will push the navigation to baseurl/search/{slug}

When I put the direct url I get more results.

search logic [handled here](https://github.com/rajesh-h/cs_content/blob/ee047a6d60b9f4d41c8a243f23e8b7f6a1e974e8/components/blocks/MainBlock.vue#L77)

I am sure I might be doing something weird, but unable to get around it.

Your help is much appreciated.

rajesh-h:
I see problem might be at where I handle $fetch hook.

when we enter direct URL it runs on server so all goes well. But on client side same effect is not observed :(

Not sure how to handle this properly

dtmzr:
Hey there, I removed this
```js
      .skip(this.skipCnt)
      .limit(this.limitCnt)
```
from your query within `MainBlock.vue`. It seemed to help and your app delivered the right content in that case.

I guess the pagination mechanism has an unexpected side effect, let me know if you need help looking into it. :)

rajesh-h:
@dtmzr Thank you very much for looking into it. Never thought this will be attended by anyone.

I think .limit() creates this issue only when you use along with .search() or .where() conditions.

Why I go with pagination mode is, lets say someone searches for word "veg" I have many tittles which has "veg" in it, so I want to show only 10 posts. so the page loads faster, also related images gets loaded only when the user wants to go down, else I might use my bandwidth of downloading 100 images which is unnecessary. Loading only when user goes down the screen saves the bandwidth in terms of users data as well.

Do you have any other way to manage this?

Thanks

dtmzr:
I looked a little further into it but sadly couldn't resolve the issue.
In your case it seems to go wrong when the client tries to fetch this data, the server seems to retrieve it correctly.

You'll notice that if you reload the page - then the correct data should be in place.

I'll try to create a minimal poc to see if this is an implementation issue.

rajesh-h:
[...]

PR #267, opened 2020-07-15
fix: order of query execution

dtmzr:
## Types of changes

## Description

Within #95 we recognized some odd behavior for the search results in the provided example.
In this issue, a possible solution was provided - it seemed like the order the in which the query is executed had some unpredicted impact on the results.
With this fix, the provided example works as expected.
The examples within this module work as expected too.

Feedback is very welcome.
If you're using advanced queries with this module, double-checking if this fix causes issues would be awesome!

## Checklist:

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -55,6 +55,29 @@ module.exports = ({ ws, database, dir, watch }) => async (req, res) => {
 
   // Build query from query / body
   let query = database.query(url, { deep: params.deep, text: params.text })
+  if (params.where) {
+    const where = {}
+
+    for (const [key, value] of Object.entries(params.where)) {
+      const [field, operator] = key.split('_')
+
+      if (operator) {
+        where[field] = {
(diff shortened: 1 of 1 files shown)


benjamincanac:
Hey @dtmzr,

Are you sure this is necessary? Since #238 has been merged, `skip` and `limit` should apply at the very end already.

This is not yet released but available on the `dev` branch, you can try it:

```json
{
  "dependencies": {
    "@nuxt/content": "nuxt/content#dev"
  }
}
```

dtmzr:
Hey @benjamincanac,

thanks for the hint, I wasn't aware of this PR.
I'll do a double-check a little later and provide feedback. 🙏

dtmzr:
@benjamincanac yep you were totally right, the with the current dev version this issue is fixed 🚀 
I'll close this PR and add this information to the issue.

Thanks for your work 💪

PR #287, opened 2020-07-23
fix: prevent options merging fn exec without default value

benjamincanac:
Functions in configuration are meant to allow merging default values. Now, functions are only executed if the value already exists in the configuration.

## Types of changes

## Description

When using `extendParser` with a function `'.txt': (file) => ({ body: file })`, the function was executed.


mathe42:
So if we want to overwrite an existing parser we have to use:

```
{
  '.xml': () => (file => parse(file))
}
```

For a new extension we can use:

```
{
  '.custom': file => parse(file)
}
```

mathe42:
Maybe we can add one sentence in the docs for this...

mathe42:
Never mind I was wrong...

PR #324, opened 2020-08-01
feat(theme-docs): allow customising default branch

danielroe:
## Types of changes

## Description
Not everyone will be preparing PRs against `dev`. It may be directly against `main` or against another branch. This PR allows customising this in settings (or automatically fetching the branch from GitHub). It defaults to `dev`, although this will be a breaking change in the (unlikely) case that users *prefer* a diff against `dev` even though they actually have a different default branch.

## Checklist:

### DIFF
--- a/docs/content/en/themes-docs.md
+++ b/docs/content/en/themes-docs.md
@@ -212,7 +212,7 @@ You can create a `content/settings.json` file to configure the theme:
   - *The GitHub repository of your project `${org}/${name}` to display the last version, the releases page, the link at the top and the `Edit this page on GitHub link` on each page*
 - `defaultBranch`
   - Type: `String`
-  - *The default branch for the GitHub repository of your project, used in the `Edit this page on GitHub link` on each page.*
+  - *The default branch for the GitHub repository of your project, used in the `Edit this page on GitHub link` on each page (defaults to `main` if it cannot be detected).*
 - `twitter`
   - Type: `String`
   - *The Twitter username you want to link*
--- a/docs/content/en/themes-docs.md
+++ b/docs/content/en/themes-docs.md
@@ -210,6 +210,9 @@ You can create a `content/settings.json` file to configure the theme:
 - `github`
   - Type: `String`
   - *The GitHub repository of your project `${org}/${name}` to display the last version, the releases page, the link at the top and the `Edit this page on GitHub link` on each page*
+- `defaultBranch`
+  - Type: `String`
+  - *The default branch for the GitHub repository of your project, used in the `Edit this page on GitHub link` on each page.*
 - `twitter`
   - Type: `String`
   - *The Twitter username you want to link*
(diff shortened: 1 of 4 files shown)


danielroe:
Note this still hard codes PRs against `docs/content` prefix, which we should be able to detect and auto-update at build time (though I think that is for another day!)

packages/theme-docs/src/store/index.js
```diff
@@ -86,6 +90,28 @@ export const actions = {
...
+      const data = await fetch(`https://api.github.com/repos/${state.settings.github}`, options).then((res) => {
+        if (!res.ok) {
+          throw new Error(res.statusText)
+        }
+        return res
+      }).then(res => res.json())
+      defaultBranch = data.default_branch
+    } catch (e) {}
+
+    commit('SET_DEFAULT_BRANCH', defaultBranch || 'dev')
```
benjamincanac, line 113: @danielroe Maybe this should be set to `master` by default?

danielroe: That would be less opinionated, although at some point that will switch to main, and so we could just start with that?

https://github.com/github/renaming

danielroe, line 113:
```suggestion
    commit('SET_DEFAULT_BRANCH', defaultBranch || 'main')
```

danielroe: let me know if you prefer `master` and I can adjust PR

docs/content/en/themes-docs.md
```diff
@@ -210,6 +210,9 @@ You can create a `content/settings.json` file to configure the theme:
 - `github`
   - Type: `String`
   - *The GitHub repository of your project `${org}/${name}` to display the last version, the releases page, the link at the top and the `Edit this page on GitHub link` on each page*
+- `defaultBranch`
```
benjamincanac, line 213: @danielroe Could you add the default value to `main` here?

packages/theme-docs/src/store/index.js
benjamincanac: `main` is fine :)

docs/content/en/themes-docs.md
```diff
@@ -210,6 +210,9 @@ You can create a `content/settings.json` file to configure the theme:
...
+  - Type: `String`
+  - *The default branch for the GitHub repository of your project, used in the `Edit this page on GitHub link` on each page.*
```
danielroe, line 215:
```suggestion
  - *The default branch for the GitHub repository of your project, used in the `Edit this page on GitHub link` on each page (defaults to `main` if it cannot be detected)*
```

docs/content/en/themes-docs.md
danielroe: :+1: Done.

benjamincanac:
Thanks @danielroe!

PR #370, opened 2020-08-07
fix(theme-docs): update spanish searching words

tincho425:
## Types of changes

## Description

Missing translation on "Searching" for Spanish. I considered "afinar" expression is not suitable.

## Checklist:

### DIFF
--- a/packages/theme-docs/src/i18n/es-ES.js
+++ b/packages/theme-docs/src/i18n/es-ES.js
@@ -1,6 +1,7 @@
 export default {
   search: {
-    placeholder: 'Buscar en documentos (Pulsa "/" para afinar)'
+    placeholder: 'Buscar en documentos (Pulsa "/" para enfocar)',
+    Searching: 'Buscando'
   },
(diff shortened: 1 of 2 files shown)


packages/theme-docs/src/i18n/es-ES.js
```diff
@@ -1,6 +1,7 @@
 export default {
   search: {
-    placeholder: 'Buscar en documentos (Pulsa "/" para afinar)'
+    placeholder: 'Buscar en documentos (Pulsa "/" para enfocar)',
+    Searching: 'Buscando'
```
adrianschubek, line 4: I would change it to lowercase `searching` as it's more consistent with other properties.

tincho425: Then apply _toUpperCase_ to the first letter when used on components? Good observation, I'll commit that change. 😉

benjamincanac:
@tincho425 Is this PR done?

PR #391, opened 2020-08-18
docs: fix dynamic routing snipet

mathieutu:
Hi,
I'm currently making a site where the routing is guided by the content files.

I made a way to achieve that, before seeing your snipet in the doc.

Howevere, I've tried to replace my piece of code by yours, but It didn't work.

The non working expected behaviour is that when you hit a folder, it should show the index file in this folder. For now, the page is just `undefined`.

Actually I don't know the difference between `$content('foo')` and `$contact().where({path: foo})`, and why we would need the `deep: true` option here, so maybe my way is not the best one to do that, but.. it works!

Thougths ?

(Ping @benjamincanac as you're the one who wrote that)

Thanks for your work!

### DIFF
--- a/docs/content/en/snippets.md
+++ b/docs/content/en/snippets.md
@@ -178,18 +178,16 @@ Let's say you want to create an app with routes following the `content/` file st
 ```vue[pages/_.vue]
 <script>
 export default {
-  async asyncData ({ $content, app, params, error }) {
-    let article
-    try {
-      [article] = await $content({ deep: true }).where({ path: `/${params.pathMatch || 'index'}` }).fetch()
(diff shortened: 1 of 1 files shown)


benjamincanac:
@mathieutu I just updated the documentation with a new version of the snippet: https://content.nuxtjs.org/snippets#dynamic-routing.

The path on the filesystem of each file is stored as is in the collection.

`$content({ deep: true }).where({ path }).fetch()` where `path` is the full route of your app allows you to bypass the default content behaviour (`$content(path)`) detecting whether it's a directory or a file to return an array or an object which prevents you to get a file with the same name as a directory.

Issue #412, opened 2020-08-25
language for installation/index page not working when navigating


debs-obrien:
### Version
@nuxt/content:  v1.7
nuxt: latest

### Reproduction Link

https://content.nuxtjs.org/fr/

### Steps to reproduce
Click link above, navigate to another page then click back to installation page when in french and it will stop working

### What is Expected?
should load index page when navigating in different lang

### What is actually happening?
won't show index page or shows not found page
<img width="1111" alt="Screenshot 2020-08-25 at 15 58 43" src="https://user-images.githubusercontent.com/13063165/91183754-21aede00-e6ec-11ea-969b-1b5b1400a065.png">
<img width="989" alt="Screenshot 2020-08-25 at 15 55 14" src="https://user-images.githubusercontent.com/13063165/91183869-499e4180-e6ec-11ea-93aa-65d5e46e486c.png">

debs-obrien:
also noticed on local generation we get this warning
 ERROR  Error generating route "/es": Page not found

PR #417, opened 2020-08-26
fix(theme-docs): handle slug index replace after localePath execution

benjamincanac:
It seems that [nuxt-i18n](https://github.com/nuxt-community/i18n-module) `localePath` method has issues with trailing slashes when not on default locale.

## Types of changes

## Description

Resolves: #412 

Since we need to remove the `/index` path, the replace was done in the `content:file:beforeInsert` hook. However when using `localePath` method on FR locale for example with `/` path, it returns `/fr` instead of `/fr/`. To solve this, a global `localeTo` method has been created to handle the `localePath` call then to replace the `/index` from the generated path.

## Checklist:

### DIFF
--- a/packages/theme-docs/src/components/global/app/AppNav.vue
+++ b/packages/theme-docs/src/components/global/app/AppNav.vue
@@ -20,7 +20,7 @@
           <ul>
             <li v-for="doc of docs" :key="doc.slug" class="text-gray-700 dark:text-gray-300">
               <NuxtLink
-                :to="localePath(doc.to).replace(/index$/, '')"
+                :to="localeTo(doc.to)"
                 class="px-2 rounded font-medium py-1 hover:text-primary-500 flex items-center justify-between"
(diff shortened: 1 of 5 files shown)


Atinux (approved):

rchl:
Oh man, please don't add another layer that messes with trailing slashes. ;) Or well, if you have too...

VueRouter does some normalization, then Nuxt and `nuxt-i18n` does also. So that would be the fourth place that does. 🙉 

Maybe if you want consistent trailing slashes you should set `router.trailingSlash` option to explicit `true` or `false` value.
And beware of bugs like https://github.com/nuxt/nuxt.js/issues/7823

rchl:
BTW. `localePath` works better with route names rather than paths. So if you pass `doc.slug` rather than `doc.to`, it might just work.

benjamincanac:
Closing this in favor of #422

PR #422, opened 2020-08-27
fix(theme-docs): extend `/` route

### LINKED ISSUE #412: language for installation/index page not working when navigating

benjamincanac:
Add an index route for prefixed locales: `/fr` 

## Types of changes

## Description

Resolves #412 

## Checklist:

### DIFF
--- a/packages/theme-docs/src/index.js
+++ b/packages/theme-docs/src/index.js
@@ -57,6 +57,16 @@ const defaultConfig = {
       // Configure `tailwind.config.js` path
       nuxt.options.tailwindcss.configPath = nuxt.options.tailwindcss.configPath || path.resolve(nuxt.options.rootDir, 'tailwind.config.js')
       nuxt.options.tailwindcss.cssPath = nuxt.options.tailwindcss.cssPath || path.resolve(nuxt.options.rootDir, nuxt.options.dir.assets, 'css', 'tailwind.css')
+      // Extend `/` route
+      nuxt.hook('build:extendRoutes', (routes) => {
+        const allRoute = routes.find(route => route.name === 'all')
+
+        routes.push({
+          ...allRoute,
+          path: '/',
+          name: 'index'
+        })
+      })
     }
   },
   content: {
--- a/packages/theme-docs/src/pages/_.vue
+++ b/packages/theme-docs/src/pages/_.vue
@@ -40,9 +40,9 @@ export default {
   layout ({ store }) {
     return store.state.settings.layout || 'default'
   },
-  middleware ({ params, redirect }) {
+  middleware ({ app, params, redirect }) {
     if (params.pathMatch === 'index') {
-      redirect('/')
+      redirect(app.localePath('/'))
     }
   },
   async asyncData ({ $content, store, app, params, error }) {


packages/theme-docs/src/index.js
```diff
@@ -57,6 +57,16 @@ const defaultConfig = {
       // Configure `tailwind.config.js` path
       nuxt.options.tailwindcss.configPath = nuxt.options.tailwindcss.configPath || path.resolve(nuxt.options.rootDir, 'tailwind.config.js')
       nuxt.options.tailwindcss.cssPath = nuxt.options.tailwindcss.cssPath || path.resolve(nuxt.options.rootDir, nuxt.options.dir.assets, 'css', 'tailwind.css')
+      // Extend `/` route
+      nuxt.hook('build:extendRoutes', (routes) => {
+        const allRoute = routes.find(route => route.name === 'all')
```
rchl, line 62: Sorry, I don't know Content code but why is that necessary. Why is the root route not there in the first place. Default `nuxt-i18n` strategy is `prefix_except_default` and it doesn't remove the unprefixed root route.

benjamincanac: It's because we use an `_.vue`, so the generated routes are:

```js
[{
  path: "/fr/*",
  component: _139e1d99,
  name: "all___fr"
}, {
  path: "/*",
  component: _139e1d99,
  name: "all___en"
}]
```

By adding an `index.vue` (done with the hook), it adds:

```js
[{
  path: "/fr",
  component: _139e1d99,
  name: "index___fr"
}, {
  path: "/",
  component: _139e1d99,
  name: "index___en"
}]

Comment (rchl) [reply packages/theme-docs/src/index.js:62]: Hmm, OK.

BTW. Not sure if related to this issue but shouldn't `localePath` be used here?:
https://github.com/nuxt/content/blob/1dff4e9e9c4bcff03396d711c37f5eee70b8690c/packages/theme-docs/src/pages/_.vue#L44-L47

Comment (rchl) [reply packages/theme-docs/src/index.js:62]: In any case, maybe just check whether this change makes both `/fr` and `/fr/` available or will server automatically redirect to `/fr/`. If both are available then it's not great for SEO.

Comment (rchl) [reply packages/theme-docs/src/index.js:62]: I'm quite confused by this bug still since the link generated for the language selector is correct.
Screenshot 2020-08-27 at 14 54 24
and things work when doing a server-side request. So it's only during generate that it fails.

I'd really like to know what is the root issue here.

Comment (benjamincanac) [reply packages/theme-docs/src/index.js:62]: Indeed, I've added the `localePath` method in the middleware. 

For SEO I don't think it's that much of an issue, although we could a canonical.

Comment (rchl) [reply packages/theme-docs/src/index.js:62]: If you are able to reproduce the issue in a minimal project (ideally without content) then I'd be happy to look at it if it's something `nuxt-i18n` related. I don't have time right now or in the near future to dig into the code of content.

Comment (Atinux) [reply packages/theme-docs/src/index.js:62]: It is not related to content directly but when using only `pages/_.vue`,  `nuxt-i18n` will create `/fr/*` and `/*` but not the `/fr` for example since there is no `pages/index.vue`

**Review by Atinux — APPROVED** [review]

File: packages/theme-docs/src/index.js (hunk 1 above)
Comment (rchl) [reply packages/theme-docs/src/index.js:62]: Yes but that is not a problem in itself as long as links are correctly created everywhere. So there is something else at play here.

Comment (Atinux) [reply packages/theme-docs/src/index.js:62]: Take a look at https://codesandbox.io/s/trusting-satoshi-rvn71?file=/pages/_.vue (I created a Sandbox especially for you).

Comment (Atinux) [reply packages/theme-docs/src/index.js:62]: You cannot hit /fr or /es, the locale will be English

Comment (rchl) [reply packages/theme-docs/src/index.js:62]: Yes but that's not unexpected technically as the catch-all route will consume everything up to a slash. When passing a literal /fr then it will be matched by the English catch-all route of course. I know that some find it unexpected but I'm on the fence on that.

Now, if all places that use `localePath` to create links would target a route name rather than passing a literal path then it would work as expected. So using crazy code like this one would work:
```js
localePath({ name: 'all', params: { pathMatch: path }})
```
Here is my fork with this change: https://codesandbox.io/s/eloquent-hawking-2ezqj

If it's expected that both /fr and /fr/ work then I think adding extra routes manually (or a middleware that redirects) is the way to go but I'm not sure if that necessarily has to work. If not handled properly it will also be bad for SEO as there will be duplicated content.

Atinux: > Now, if all places that use localePath to create links would target a route name rather than passing a literal path then it would work as expected.
This should work without giving the name (since they don't necessary know it since Nuxt generates them).

> If it's expected that both /fr and /fr/ work then I think adding extra routes manually
Yes that's the point of this PR actually, this is why we add the `index` page 😉 

> If not handled properly it will also be bad for SEO as there will be duplicated content.
Well actually, crawlers are not that dumb if the content is the same with or without trailing slash, see https://www.google.com/search?q=nuxt+content+site%3Acontent.nuxtjs.org as we don't necessary forces a trailing slash.

Next step is to add a canonical meta tag for the theme to specify the preferred route (I like without the trailing slash personally)

rchl: I'm discussing this from the bigger perspective of handling this in `nuxt-i18n`. I'm not arguing against this fix or anything.

I'd say that what you are saying "should" happen is technically less correct as someone, for some reason, might want to have both /fr and /fr/ working and point to different things (when using catch-all routes the /fr might actually refer to some separate content if everything is dynamic).
It's probably an edge case through an in a typical case user would maybe expect both to go to a French index.

And I'm not sure if handling this on the `nuxt-i18n` level would the right thing to do.

rchl: Well, maybe a super-specific fix like "if localePath argument starts with a slash and matches locale exactly then go the the route matching that locale" could be implemented but then I'll be potentially shadowing user's dynamic route handler so in those edge cases someone might complain.

Atinux: Indeed I don't think it should be handled right inside nuxt-i18n, I think this fix here makes sense :)

PR #428, opened 2020-08-27
fix(theme-docs): improve live edit styles

ThomOrlo:
## Types of changes

- [X] Bug fix (a non-breaking change which fixes an issue) → **dark mode**
- [X] New feature (a non-breaking change which adds functionality) → **light mode**

## Description

Before (light mode):
before-live-edit-light

Before (dark mode):
before-live-edit-dark

After (light mode):
after-live-edit-light

After (dark mode):
after-live-edit-dark

## Checklist:

### DIFF
--- a/packages/theme-docs/src/assets/css/main.css
+++ b/packages/theme-docs/src/assets/css/main.css
@@ -82,3 +82,11 @@
     @apply bg-gray-800 static;
   }
 }
+
+.nuxt-content-editor {
+  @apply rounded-md bg-gray-200;
+}
+
+.dark-mode .nuxt-content-editor {
+  @apply bg-gray-800;
+}
--- a/packages/theme-docs/src/assets/css/main.css
+++ b/packages/theme-docs/src/assets/css/main.css
@@ -82,11 +82,3 @@
     @apply bg-gray-800 static;
   }
 }
-
-.nuxt-content-editor {
-  @apply rounded-md bg-gray-200;
-}
-
-.dark-mode .nuxt-content-editor {
-  @apply bg-gray-800;
-}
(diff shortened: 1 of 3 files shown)


mathe42:
Great PR. I think you should add this styling directly in the editor.vue file. Your solution adds not used CSS to production build.

ThomOrlo:
If I add this styling directly in the `editor.vue` file or` nuxt-content.dev.vue` file, it will be for everyone who uses `@nuxt/content`.
My proposal is just for `@nuxt/content-theme-docs`.

benjamincanac:
@ThomOrlo It looks really nice! 
@mathe42 You're right, this shouldn't be defined in the `main.css` but it shouldn't be defined in `editor.vue` either, we can't use TailwindCSS classes in @nuxt/content. Do you have a suggestion on extending `nuxt-content-dev` component styles?

Atinux:
I think it would be nice to use directly a custom editor for the docs theme directly (thanks to v1.7.0) so it could directly be used with custom css for it

ThomOrlo:
Sorry I can't understand everything 😕 

From what I understood, `main.css` is injected globally:
https://github.com/nuxt/content/blob/dev/packages/theme-docs/src/index.js#L23
```js
css: [
  '~/assets/css/main.css'
],
```

For exemple `.filename`:
```css
& > .filename {
  @apply absolute right-0 top-0 text-gray-100 z-10 italic leading-none mr-3 mt-3;
}
```

These styles are visible on the final site :
explanations-main-style

With `.nuxt-content-editor`, it will work too, right?

Atinux:
One way could be to add a main.dev.css and push it in development from the module to avoid including the css for editor or directly add the css in the nuxt-content.dev.js component

PR #429, opened 2020-08-27
fix(content): Offline mode

### LINKED ISSUE #229: Content And Offline mode
Hello,

I'm trying to build an web app using this module and the [pwa module](https://pwa.nuxtjs.org/). When I try to go offline, the navigator says that he can't fetch the database. Is this a bug, or I miss something or is this impossible ?

image

Barbapapazes:
Fix issue when using app in SPA mode. The DB wasn't call correctly (using a `Date.now()` which makes unusable offline mode)

## Types of changes

## Description

It resolve #229 ! Using a env in the runtimeConfig and passing it to the static plugin, the value is the same during the app is alive but change when there is a new build !

Maybe I can use a key generation but don't kwon how !

## Checklist:

### DIFF
--- a/packages/content/lib/index.js
+++ b/packages/content/lib/index.js
@@ -113,7 +113,10 @@ module.exports = async function (moduleOptions) {
       }
     })
 
-    const path = paths.join('/').replace(/\/+/g, '/').replace(/^\//, '')
+    const path = paths
+      .join('/')
+      .replace(/\/+/g, '/')
+      .replace(/^\//, '')
 
     return database.query(`/${path}`, options)
   }
@@ -159,7 +162,7 @@ module.exports = async function (moduleOptions) {
     // Create a hash to fetch the database
     const dbHash = hash(JSON.stringify(database.items._data)).substr(0, 8)
     // Pass the hash to the publicRuntimeConfig to be used in client side
-    this.options.publicRuntimeConfig.content = dbHash
+    this.options.publicRuntimeConfig.content = { dbHash }
     // Write db.json
     this.nuxt.hook('generate:distRemoved', async () => {
       const dir = resolve(this.options.buildDir, 'dist', 'client', 'content')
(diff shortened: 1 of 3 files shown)


benjamincanac:
@Barbapapazes Really nice PR! Just need to fix the tests and it will be good to go 🙂

Barbapapazes:
Tests is ok !

benjamincanac (approved):

packages/content/lib/index.js
```diff
@@ -138,12 +156,20 @@ module.exports = async function (moduleOptions) {
 
   /* istanbul ignore if */
   if (isSSG) {
+    // Create a hash to fetch the database
+    const dbHash = hash(JSON.stringify(database.items._data)).substr(0, 8)
```
pi0, line 160: Is it okay that not other database changes invalidate hash?

packages/content/lib/index.js
```diff
@@ -138,12 +156,20 @@ module.exports = async function (moduleOptions) {
...
+    // Pass the hash to the publicRuntimeConfig to be used in client side
+    this.options.publicRuntimeConfig._CONTENT_HASH_DB = dbHash
```
pi0, line 162: runtimeConfig is a js object not env to be uppercase. We may use `.publicRuntimeConfig.content = { dbHash }`

pi0 (commented):
Nice improvement!

packages/content/lib/index.js
Atinux: Agree :)

packages/content/lib/index.js
Atinux: Yes since it's only stored in items, right @benjamincanac ?

Barbapapazes: I can use the entire database file to generate the hash if you think it's better but I analyse the file and only the field items._data contains the data !

packages/content/lib/index.js
Barbapapazes: ok, so I change this !

Atinux (approved):
Awesome work 🔥

PR #435, opened 2020-08-30
fix slug for index.md files

### LINKED ISSUE #385: slug property is not correct when using directory and index.md approach
### Version
@nuxt/content: 1.6.0
nuxt: 2.14.1

### Reproduction Link
https://codesandbox.io/s/dank-paper-djlm0?file=/content/a-new-entry/index.md

### Steps to reproduce
Create two content files
 - one with a unique name at the top level in the content directory
 - one in a subdirectory, and name it index.md

Start the application and check the content. Supply the file name or the directory name as the slug in ` await $content(slug)`

### What is Expected?
the slug property on the returned post is the same as what had to be supplied in order to fetch it

the slug property is correct for file names, but is being returned as 'index' for directory based content ie. the .md file name not the actual slug name

azrikahar:
## Types of changes

## Description

Resolves #385.

This change is to fix the slug for nested `index.md` files, which means all the index.md files will have the same slug. For example:

- Current slug
  - path: `/a-new-entry/index.md`
  - slug: `index`
- Proposed slug
  - path: `/a-new-entry/index.md`
  - slug: `a-new-entry`

Here are the screenshots to show comparison with more test cases, especially the slug for `/a-new-entry/index.md`.

| Original | Proposed |
| --- | --- |
| chrome_hBLQ0HnkNy | chrome_g5ltSGDqgN |

## Checklist:

Not sure whether this requires test as I'm not very clear on this area. Hope to receive input on this.

### DIFF
--- a/packages/content/lib/database.js
+++ b/packages/content/lib/database.js
@@ -207,7 +207,7 @@ class Database extends Hookable {
       dir: dir || '/',
       path: normalizedPath,
       extension,
-      slug: split[split.length - 1],
+      slug: /\/+.*\/index$/.test(normalizedPath) ? split[split.length - 2] : split[split.length - 1],
       createdAt: isValidDate(existingCreatedAt) ? existingCreatedAt : stats.birthtime,
       updatedAt: isValidDate(existingUpdatedAt) ? existingUpdatedAt : stats.mtime
     }


Atinux:
Hi @azrikahar 

What happens if I have `a-new-entry.md` file then?

azrikahar:
> Hi @azrikahar
> 
> What happens if I have `a-new-entry.md` file then?

That's a very good question. I'm sure in this case, `/a-new-entry/index.md` and `/a-new-entry/a-new-entry.md` will end up having the same slug. 

On hindsight, even with both the current implementation (without regex) and my proposed fix, `/path-a/a-new-entry.md` and `/path-b/a-new-entry.md` will still end up with the same slug. Seems like we are still hitting a wall here.

This just occured to me after you asked this question, but I think that's why sites like [dev.to](https://dev.to) and [medium.com](https://medium.com) always have hashes at the end of the slugs. Maybe that will be the only correct solution here. However we'll have to make sure the hash is based on the path and not the slug only so the hash will be preserved regardless how many times we generate them, while still remaining different for `same slug different path` scenarios.

I'll turn this PR to a draft first for now, but I'm also okay with closing it if it's not a priority atm.

midorikocak:
What happened to this?

Issue #432, opened 2020-08-28
Add the option to extend default parser functionality


nielsdB97:
### Is your feature request related to a problem? Please describe.

I would like to extend the markdown parser to strip my `// prettier-ignore` comments. These are needed in development but I don't want my visitors to see them. I have found the [`extendParsers`](https://content.nuxtjs.org/configuration#extendparser) option but as far as I can tell it will [override](https://github.com/nuxt/content/blob/4b3417edde4faed1cc92153881e2527575d3e28e/packages/content/lib/database.js#L181) the original parser.

### Describe the solution you'd like

Instead of overriding the original parser, I would like to extend its functionality by mutating the string it will parse.

### Describe alternatives you've considered

Re-implement the default parser.

### Additional context

This is the logic I came up with for stripping prettier comments:
```typescript
function removePrettierIgnoreFromCodeBlocks(file: string) {
	const fileWithoutPrettierComments = file
		.replace(/^ \/\/ prettier-ignore\n/gm, "")
		.replace(/ \/\/ prettier-ignore\n/gm, "\n");

	return fileWithoutPrettierComments;
}
```

mathe42:
You can overwrite the paser with something like (untestet!):

```js
const md = new require('@nuxt/content/parser').Markdown(options.markdown /* add default options combined with your own options */)

const myParser = (file) => md.toJSON(removePrettierIgnoreFromCodeBlocks(file))
```

nielsdB97:
@mathe42 thank you so much for the quick response! Your solution works perfectly, the only downside is that I have to use internals for which there are no types available. I therefore have to stick to `require` statements rather than using `import`.

I still wonder if it would be nice to have a function-like syntax for extending the config without having to use these internals, something like this:

```javascript
function customMarkdownParser(file: string) {
	const MarkdownParser = require("@nuxt/content/parsers").Markdown;
	const { getOptions } = require("@nuxt/content/lib");

	const markdownOptions = getOptions().markdown;
	const md = new MarkdownParser(markdownOptions);

	const fileWithoutPrettierIgnore = removePrettierIgnoreFromCodeBlocks(file);
	return md.toJSON(fileWithoutPrettierIgnore);
}

export default {
	content: {
		extendParser: (defaultParser) => ({
			...defaultParser,
			".md": defaultParser.md(customMarkdownParser),
		}),
	},
}
```

For now I ended up with this config:

```typescript
function removePrettierIgnoreFromCodeBlocks(file: string) {
	return file
[...]

PR #450, opened 2020-09-02
feat: add better parser extension

mathe42:
## Types of changes

## Description

`extendParsers` can be function:
```ts
export default {
	content: {
		extendParser: (defaultParser) => ({
			".md": file => postCompiler(defaultParser[".md"](preCompiler(file))),
		}),
	},
}
```

see #432

## Checklist:

### DIFF
--- a/docs/content/en/configuration.md
+++ b/docs/content/en/configuration.md
@@ -319,7 +319,7 @@ This module uses `node-csvtojson` to parse csv files, you can check here for [op
 
 ### `extendParser`
 
-- Type: `Object`
+- Type: `Object | Function`
 - Default `{}`
 
 With this option you can define your own parsers for other file types. Also you can **overwrite** the default parser!
@@ -339,6 +339,25 @@ const parseTxt = file => file.split('\n').map(line => line.trim())
   }
 }
 ```
+
+#### Manipulating existing parsers
+You can change the default parsers! You get them when you provide a function:
+
+**Example**
+
+```js{}[nuxt.config.js]
+const parseTxt = file => file.split('\n').map(line => line.trim())
+
+// in Config:
+
+{
+  extendParser: (defaultParser) => ({
+	  ".md": file => postCompiler(defaultParser[".md"](preCompiler(file))),
+	})
+}
+```
+
+
 
 ### `editor`
 
--- a/packages/content/types/index.d.ts
+++ b/packages/content/types/index.d.ts
@@ -31,7 +31,7 @@ declare module '@nuxt/vue-app' {
 
 type extendOrOverwrite<T> = ((old: T) => T) | T
 
-type parsers = { [extension: string]: (file: string) => any }        
+type parsers<T extends string> = { [extension: T]: (file: string) => any }        
         
 // Nuxt 2.9+
 declare module '@nuxt/types' {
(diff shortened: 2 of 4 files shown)


mathe42:
@nielsdB97 can you have a look at this syntax?

Atinux:
I am thinking that adding a hook names “content:file:beforeParse” might to the trick for this

docs/content/en/configuration.md
```diff
@@ -340,6 +340,25 @@ const parseTxt = file => file.split('\n').map(line => line.trim())
...
+**Example**
+
+```js{}[nuxt.config.js]
+const parseTxt = file => file.split('\n').map(line => line.trim())
+
+// in Config:
+
+{
+  extendParser: (defaultParser) => ({
+	  ".md": file => postCompiler(defaultParser[".md"](preCompiler(file))),
```
nielsdB97, line 355: It's not entirely clear to me where `postCompiler` and `preCompiler` are coming from here. I suppose these are the custom functions that the user is supposed to define theirselves?

packages/content/types/index.d.ts
```diff
@@ -55,9 +57,7 @@ declare module '@nuxt/types' {
       yaml?: yamlOptions,
       csv?: csvOptions,
       xml?: xmlOptions,
-      extendParser?: {
-        [extension: string]: (file: string) => any
-      }
+      extendParser?: parsers<string> | (defaultParsers: parsers<'.md'|'.yml'|'.yaml'|'.json'|'.json5'|'.xml'>) => parsers<string>
```
nielsdB97, line 60: Can we allow users to extend the `parsers<'.md'|'.yml'|'.yaml'|'.json'|'.json5'|'.xml'>` list with their own extensions? As to extend the list to match their config, maybe by passing a generic parameter?

mathe42: The default parsers have known extensions so no need to extend this.

docs/content/en/configuration.md
mathe42: Yes they are defined by the user.

packages/content/types/index.d.ts
nielsdB97: This entire PR is to allow users to define their own extensions. So I mean to let them add to this list.

mathe42: These are the default parsers so only for files with extension md, yml, yaml, json, json5, xml. The user can not change what he gets 

See in Code:
https://github.com/nuxt/content/blob/95d2712de1f5e86e56c77c21e08a108df596f2a7/packages/content/lib/database.js#L26-L34
https://github.com/nuxt/content/blob/95d2712de1f5e86e56c77c21e08a108df596f2a7/packages/content/lib/database.js#L44

mathe42: Oh missed csv.

nielsdB97: Oh my bad, it's the argument passed to the user of course... Nevermind then, looks good 👍

docs/content/en/configuration.md
nielsdB97: Can we make this more clear by adding an empty function declaration above or something like that?

```typescript
function preCompiler(file: string) {}
function postCompiler(parsedFile: any) {}

{
  extendParser: (defaultParser) => ({
    ".md": file => postCompiler(defaultParser[".md"](preCompiler(file))),
  })
}
```

nielsdB97: Also maybe we should call them `preParser` since that term is used throughout this functionality, rather than `preCompiler`.

mathe42: Yes I'm not happy with the docs. I will work on that later if you want you can write something :D

Atinux:
Closing in favour of https://github.com/nuxt/content/pull/453

Issue #439, opened 2020-08-31
Support Custom Domain or API Root from which to fetch repository data


NozomuIkuta:
### Is your feature request related to a problem? Please describe.

I want to use the Doc Theme for user guide of internal service of my company, which is hosted in GitHub Enterprise.

But, currently, I can fetch data from GitHub only, since the URL is statically defined in the source code ([here](https://github.com/nuxt/content/blob/05c22447c3063593c0dbbc34eeee3e09bee51f90/packages/theme-docs/src/store/index.js#L67), and [here](https://github.com/nuxt/content/blob/05c22447c3063593c0dbbc34eeee3e09bee51f90/packages/theme-docs/src/store/index.js#L105)).

### Describe the solution you'd like

It would be nice if developers can override the URL from which they fetch data, so that they don't have to folk and set up the theme from scratch.

```json
// settings.json
{
  "github": "https://custom.com/path/to/"
}
```

```javascript
// src/store/index.js
try {
  const url = state.settings.github.startsWith('http')
    ? state.settings.github
    : `https://api.github.com/repos/${state.settings.github}`
  const data = await fetch(url, options).then((res) => { ... })
} catch (e) { }

// and...

try {
  const url = state.settings.github.startsWith('http') 
    ? `${state.settings.github}/releases`
    : `https://api.github.com/repos/${state.settings.github}/releases`
  const data = await fetch(url, options).then((res) => { ... })
} catch (e) { }
```

### Describe alternatives you've considered

It might be an option to let developers define `github` object where keys are `releases` and `defaultBranch`, and values are their full urls.

### Additional context

Prompt which is shown when I run `yarn create nuxt-content-docs <project-name>` would be affected, or can be kept as what it is.

PR #458, opened 2020-09-05
feat: support github custom domain

NozomuIkuta:
## Types of changes

## Description

This PR implements #439, and lets us fetch repositories hosted on GitHub Enterprise, whose API has other domain than `api.github.com`.

I defined a Vuex getter `githubUrls` that returns an object whose values are full urls to get data of repository and its releases.

In addition, I modified related part of docs and SAO prompts.

## Checklist:

#### `create-nuxt-content`

First, I added a test case to check whether it creates files correctly when developer assign full url of repository instead of `owner/repo` string.

Second, I ran `yarn test` in the project root and all the tests were passed.

#### `theme-docs` 

First, I ran `yarn link` for all the packages (`@nuxt/content`, `@nuxt/theme-docs`, and `create-nuxt-content`). 

Second, I ran the linked `create-nuxt-content` to get `theme-docs` files. 

Third, I replaced the installed `@nuxt/theme-docs` with my local one, and ran `yarn run dev`, resulting in that releases are correctly fetched.

Screen Shot 2020-09-06 at 0 15 54

It would be fine to test the full-url case with GitHub repository instead of GitHub Enterprise, because what is different is only the form of the original string, not actual reachability of the API.

---

P.S. This is my very first PR for open source project in my life. 👶

### DIFF
--- a/packages/theme-docs/src/store/index.js
+++ b/packages/theme-docs/src/store/index.js
@@ -17,11 +17,27 @@ export const getters = {
     return state.settings
   },
   githubUrls (state) {
-    const { github } = state.settings
+    const { github = '', githubApi = '' } = state.settings
 
-    return github.startsWith('http')
-      ? { repo: github, releases: `${github}/releases` }
-      : { repo: `https://api.github.com/repos/${github}`, releases: `https://api.github.com/repos/${github}/releases` }
+    // GitHub Enterprise
+    if (github.startsWith('http') && githubApi.startsWith('http')) {
+      return {
+        repo: github,
+        api: {
+          repo: githubApi,
+          releases: `${githubApi}/releases`
+        }
+      }
+    }
+
+    // GitHub
+    return {
+      repo: `https://github.com/repos/${github}`,
+      api: {
+        repo: `https://api.github.com/repos/${github}`,
+        releases: `https://api.github.com/repos/${github}/releases`
+      }
+    }
   },
   releases (state) {
     return state.releases
@@ -72,7 +88,7 @@ export const actions = {
     }
     let releases = []
     try {
-      const data = await fetch(getters.githubUrls.releases, options).then((res) => {
+      const data = await fetch(getters.githubUrls.api.releases, options).then((res) => {
         if (!res.ok) {
           throw new Error(res.statusText)
         }
@@ -110,7 +126,7 @@ export const actions = {
     }
     let defaultBranch
     try {
-      const data = await fetch(getters.githubUrls.repo, options).then((res) => {
+      const data = await fetch(getters.githubUrls.api.repo, options).then((res) => {
         if (!res.ok) {
           throw new Error(res.statusText)
         }
--- a/packages/theme-docs/src/store/index.js
+++ b/packages/theme-docs/src/store/index.js
@@ -16,6 +16,13 @@ export const getters = {
   settings (state) {
     return state.settings
   },
+  githubUrls (state) {
+    const { github } = state.settings
+
+    return github.startsWith('http')
+      ? { repo: github, releases: `${github}/releases` }
+      : { repo: `https://api.github.com/repos/${github}`, releases: `https://api.github.com/repos/${github}/releases` }
+  },
   releases (state) {
     return state.releases
   },
@@ -54,7 +61,7 @@ export const actions = {
 
     commit('SET_CATEGORIES', categories)
   },
-  async fetchReleases ({ commit, state }) {
+  async fetchReleases ({ commit, state, getters }) {
     if (!state.settings.github) {
       return
     }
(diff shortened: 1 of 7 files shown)


andoshin11:
LGTM

Atinux (approved):
Awesome work done @NozomuIkuta 👏

NozomuIkuta:
@Atinux 

Thank you for reviewing!

In fact, I found some bugs in my PR. 🙇 

That is, `state.settings.github`, which now can be either `"owner/repo"` or full url, is still used in 3 components ([here](https://github.com/nuxt/content/blob/162d54b8074b4777434f213e90ddc47300b6b27a/packages/theme-docs/src/components/app/AppHeader.vue#L55), [here](https://github.com/nuxt/content/blob/162d54b8074b4777434f213e90ddc47300b6b27a/packages/theme-docs/src/components/app/AppGithubLink.vue#L35), and [here](https://github.com/nuxt/content/blob/162d54b8074b4777434f213e90ddc47300b6b27a/packages/theme-docs/src/components/app/AppNav.vue#L53)). I will modify these parts for them to handle `getters.githubUrls` object instead of `state.settings.github`.

NozomuIkuta:
Sorry for multi-posting. I found a thing to reconsider.

According to [GitHub Enterprise's spec](https://developer.github.com/enterprise/2.21/v3/enterprise-admin/), it is possible that host name for repositories and that for API are different. As far as I googled, there are some patterns like following:

| Normal Root                         | API Root                                                | To compute `getters.githubUrls` ...                                  |
| ------------------------ |-------------------------------- | --------------------------------------------------  |
| http://github.com               | https://api.github.com                     | Only `"owner/repo"` is  necessary                                    |
| http://my-company.com | https://api.my-company.com        | 2 full urls to the repo and api are necessary                    |
| http://my-company.com | https://my-company.com/api/v3 | 2 full urls to the repo and api are necessary                    |

Fortunately, it seems that the path to edit files and that to get releases are always same.

- Edit: `path/to/repo/edit/branch/filename`
- Releases: `path/to/repo/releases`

---

@Atinux 

Then, I have a question about how to handle this issue.

Which do you think is approach that is most suitable for Nuxt's way or its DX?

**Case 1**

- Add a prompt to `create-nuxt-content` for it to ask 2 urls, url for a repo and api-url for the repo.
  - In the case of GitHub, user can answer like `owner/repo` and skip the api-url.
  - In the case of GitHub Enterprise, user is promoted to answer 2 full urls.
  - In both cases, user can edit `content/settings.json` later.

**Case 2**

- Revert the prompt that asks GitHub settings, so that it appears to ask a string in the form of `"owner/repo"`.
  - In the case of GitHub Enterprise, user has to edit `content/settings.json` later.
  - In both cases, user can edit `content/settings.json` later.

Whichever, `getters.githubUrls` provides an object of urls, so all related UIs can render the urls correctly.

Atinux:
@NozomuIkuta I prefer the Case 2, for advanced use case, they can edit settings.json, this way we keep as simple as simple the create-nuxt-content-docs :)

NozomuIkuta:
@Atinux 

I fixed my PR based on your opinion.

First, I reverted changes on `create-nuxt-content-docs` so it asks `owner/repo` string to cover a basic use case.

Second, I added `githubApi` setting for developer to define API full url besides `github` setting (cf. [my comment above](https://github.com/nuxt/content/pull/458#issuecomment-689028881)).

An alternative approach would be to make `github` setting able to be defined as a string, or an object which has 2 full urls. Since I though it simpler, I chose adding a setting over changing "type" of an existing one. But, I'm not sure which is more suitable so let me know if you have an opinion.

Third, I modified related parts of docs again (adding description about `githubApi` setting).

Forth, of course I ran `yarn run test` after development, and all the tests were passed :)

packages/theme-docs/src/store/index.js
```diff
@@ -15,6 +15,29 @@ export const getters = {
   settings (state) {
     return state.settings
   },
+  githubUrls (state) {
+    const { github = '', githubApi = '' } = state.settings
+
+    // GitHub Enterprise
+    if (github.startsWith('http') && githubApi.startsWith('http')) {
```
NozomuIkuta, line 22: Is it enough to check just existence of the 2 settings?

```javascript
 if (github && githubApi) { ... }
```

NozomuIkuta:
I noticed that this PR will have conflict with #468 in `link` computed property of AppGithubLink component. IMHO, it would be safer to merge #468 in advance so that I can rebase my branch and replace `this.settings.github` with `this.githubUrls.repo`.

benjamincanac:
@NozomuIkuta Oh yes indeed, #468 has been merged!

NozomuIkuta:
@benjamincanac

Thank you for letting me know!
I rebased my branch, resolved conflicts, and pushed it again. 🙋‍♂️

packages/theme-docs/src/store/index.js
benjamincanac: It should be enough I guess

benjamincanac (approved):

Atinux (approved):

PR #472, opened 2020-09-12
fix(theme-docs): fix code block's filename style

### LINKED ISSUE #463: Add padding to filename in codeblocks
It's a minor one. When you include the filename to a codeblock, it touches the code below when displayed.

A `padding-bottom: 2px;` or similar solution will do!

None

### Additional context
Screen Shot 2020-09-09 at 15 37 52

azrikahar:
## Types of changes

## Description

Fixes #463.

### Before Fix

As the issue highlighted, the filename will intersect with the first line of a code block. Below is a sample screenshot, one with no filename and the one with filename exhibits this issue:

_Note: They may not look super intersecting like the one show in the issue, but it's because only characters with long tail like "q", "p" or "g" are the worst offender._

chrome_7QSL6WFJ9V

### After Fix

The obvious fix is to add a top padding to code blocks, but this will make code blocks without filename look awkward with top gap. This is the reason why I used the `+` operator in `.filename + pre[class*="language-"]` which **will only target the `pre` tag if there is a div with `.filename` class above it.** The resulting fix is shown in the screenshot below:

chrome_AOAuNRLwqp

Notice how the top code block has equal top and bottom padding like usual, but only the bottom code block with file name has a larger top padding to compensate.

### Slight out-of-topic fix

Just in case you guys are wondering why I changed the styling for `.filename` itself, currently the filename is styled with equal top and right margin to space it away from the top right corner, as indicated by the `mt-3` and `mr-3` here:

https://github.com/nuxt/content/blob/a72d92e324a104f8763a67bd7953976358da7b5e/packages/theme-docs/src/assets/css/main.css#L57

However from my experience, typically we need more padding on the X axis (left & right) compared to the Y axis (top & bottom) when it comes to words. Here are my reasoning for my change on that line:

|Before ( mr-3 ) | After ( mr-4 } |
|---|---|
| chrome_aKc5Q5CUMO | chrome_2BZ67UkrjV |

In the example above, we can see that even though originally the top and right margin is the same, it still seems like the right margin is visually less than the top margin. This is why we need to compensate more for the right margin.

Just to add on another example from the Nuxt official website's subscribe button for newsletter:

| Normal Visual | Padding shown via DevTools |
|---|---|
| chrome_JvH7DpRaP2 | ShareX_JDE5foyeTh |

This once again shows that even when the word `SUBSCRIBE` visually has same padding from all sides, but actually the left and right padding is more than top and bottom padding. Sorry if I'm rambling on too much for this little thing haha.

## Checklist:

### DIFF
--- a/packages/theme-docs/src/assets/css/main.css
+++ b/packages/theme-docs/src/assets/css/main.css
@@ -54,7 +54,12 @@
 
     /* Style filename span added by @nuxt/content */
     & > .filename {
-      @apply absolute right-0 top-0 text-gray-100 z-10 italic leading-none mr-3 mt-3;
+      @apply absolute right-0 top-0 text-gray-100 z-10 italic leading-none mr-4 mt-3;
+    }
+
+    /* Add top padding to code blocks with filename */
+    & > .filename + pre[class*="language-"] {
+      @apply pt-8;
     }
 
     /* Style copy button added in `pages/_.vue` */
--- a/packages/theme-docs/src/assets/css/main.css
+++ b/packages/theme-docs/src/assets/css/main.css
@@ -54,7 +54,7 @@
 
     /* Style filename span added by @nuxt/content */
     & > .filename {
-      @apply absolute right-0 top-0 text-gray-100 z-10 italic leading-none mr-4 mt-3;
+      @apply absolute right-0 top-0 text-gray-100 z-10 font-mono text-sm tracking-tight leading-none mr-4 mt-3;
     }
 
     /* Add top padding to code blocks with filename */


Atinux:
This is just amazing to review your PRs @azrikahar

Atinux (approved):

benjamincanac:
Hey @azrikahar, awesome work! I just improved the typography a bit, what do you think?

benjamincanac (approved):

azrikahar:
> This is just amazing to review your PRs @azrikahar

@Atinux thank you very much! It really means a lot coming from you. 

> Hey @azrikahar, awesome work! I just improved the typography a bit, what do you think?
> 
> Screenshot 2020-09-14 at 12 06 11

@benjamincanac looks great! The tracking definitely makes it more readable while the font-mono takes the role of the italic which is making it unique from the code lines. Nice touch :)

PR #502, opened 2020-09-24
PROPOSAL: Make other file types to also expose their text source.

renoirb:
**Idea** Make other file types to also expose their text source.

## Types of changes

- New feature (a non-breaking change which adds functionality)

## Description

In some situations, we might want to have the raw source to use with own syntax highlighter, say for a static site.

The `$content([...], { text: true })` only works for Markdown. This should allow for other types too.

## Checklist:


packages/content/lib/database.js
```diff
@@ -200,7 +200,12 @@ class Database extends Hookable {
     // Collect data from file
     let data = []
     try {
-      data = await parser(file.data)
+      const text = await file.data
```
renoirb, line 203: That's assuming data is an `async` getter and not the parser that's the asynchronous.

benjamincanac:
@renoirb What is the use-case here? If you want to do this, it would be best to return the text content from the parsers, see how it's done for markdown: https://github.com/nuxt/content/blob/dev/packages/content/parsers/markdown/index.js#L107.

renoirb:
> @renoirb What is the use-case here? 

The use-case I had is when I want to display the raw source. The markdown source is available in raw probably for the editor. In my case, I wanted the raw YAML source so that I could show it in a colorized code block.

renoirb:
Hey @benjamincanac do you want me to clarify the use-case?

renoirb:
FC816AE4-D442-4985-90CE-DA9BA4675A6B

Tahul:
Hey @renoirb 😄

I think you could achieve the exact result from your screenshot by using code block in @nuxt/content v2.

YML Syntax highlighting will be provided by [Shiki](https://github.com/shikijs/shiki)!

I am closing this in favor of v2 but feel free to re-open if you want to continue the work on this PR.

renoirb:
I shall have a look. I can't wait to see how nuxt content changed.

PR #507, opened 2020-09-26
feat(theme-docs): support `docs.primaryColor`

pi0:
## Types of changes

## Description

With this new feature, users can easily define `primaryColor` in `nuxt.config` like this:

```js
import theme from '@nuxt/content-theme-docs'
  
export default theme({
  docs: {
    primaryColor: '#E24F55'
  }
})
```

This extends default `tailwind.config.js` by generated colors using [nuxt-contrib/theme-clors](https://github.com/nuxt-contrib/theme-colors) which is based on [tailwindcolorshades](https://javisperez.github.io/tailwindcolorshades) weights.

## Checklist:

- [x] Manually test functionality
- [x] Update docs and also migration guide
- [x] Use for `loading-color` and `theme_color`


Atinux:
I think we could also set the `loading.color` property and also `pwa.manifest.theme_color`

benjamincanac (commented):
@pi0 Just to be sure, when upgrading, will it still work for people who defined primary colors in their `tailwind.config.js`?

pi0:
@benjamincanac Yep ;) I double checked with an external repo. I think we should provide guide to migrate using new config since we are applying default color for loading.color and pwa now.

benjamincanac:
@pi0 Where would you put the migration guide? I don't even know where to put release notes for the theme when using lerna.

pi0:
Maybe in github release notes? :) I've moved default primaryColor to tailwind config to avoid changing `loading` color unless specified. This way we don't need migration notes too (however recommended)

benjamincanac (approved):

Atinux (approved):

PR #512, opened 2020-09-28
docs: describe `catch` in fetching methods

samyrooster:
Added catch() after fetch() to prevent error.

## Types of changes

## Checklist:


benjamincanac:
@samyrooster I think it would be best to put this in a snippet "Error catching" to demonstrate with `.catch` and `try {} catch () {}` since it's not really related to `@nuxt/content`.

samyrooster:
@benjamincanac try {} catch () doesn't work on nested pages (`<nuxt-child />`).
And also on example page (https://content.nuxtjs.org/examples/basic) used `catch()` which doesn't mentioned on documentions.

PR #533, opened 2020-10-03
feat(database): add metadata for created and updated from git logs

andrewkreuzer:
## Types of changes

## Description

Adds createdAt and updatedAt metadata from git logs. Taking the date from the latest commit for a file and the first commit for a file to get the gitUpdatedAt and gitCreatedAt metadata respectively.

When using autodeploy solutions (Netlify) which clone a repo containing a nuxtjs/content site the metadata for updatedAt and createdAt are lost

## Checklist:

- [X] My change requires a change to the documentation.


benjamincanac:
Hi @andrewkreuzer,

Thank you for your PR, unfortunately I don't think this should live in the core of content. You can achieve the same result by using the `beforeInsert` hook.

andrewkreuzer:
Ah good point, thanks for the feedback that would be a better solution.

dword-design:
@andrewkreuzer @benjamincanac I built an [npm package](https://github.com/dword-design/nuxt-content-hooks-git) that basically does the thing.

By default it replaces `createdAt` and `updatedAt`, but you can also configure the fields.

Install it: `npm install nuxt-content-hooks-git`

Then add it to your Nuxt.js config like this:

```js
import nuxtContentHooksGit from 'nuxt-content-hooks-git'

export default {
  modules: ['@nuxt/content'],
  hooks: {
    // other hooks
    ...nuxtContentHooksGit(),
  },
}
```

Or configure the fields:

```js
import nuxtContentHooksGit from 'nuxt-content-hooks-git'

export default {
  modules: ['@nuxt/content'],
  hooks: {
    // other hooks
    ...nuxtContentHooksGit({ createdAtName: 'gitCreatedAt', updatedAtName: 'gitUpdatedAt' }),
  },
}
```

andrewkreuzer:
Oh nice! Might have to switch to this, this'll clean up my nuxt.config.js! thanks! @dword-design

dword-design:
@andrewkreuzer @benjamincanac I actually realized that it is much cleaner to make a Nuxt.js module out of it because it's easier to pass options to it and it does not interfere with other hooks (e.g. someone wants to implement another `beforeInsert` hook. So here is the module alternative:

```bash
npm install nuxt-content-git
```

```js
// nuxt.config.js

export default {
  modules: [
    'nuxt-content-git',
    '@nuxt/content',
  ]
}
```

Or with options:
```js
export default {
  modules: [
    ['nuxt-content-git', { createdAtName: 'gitCreatedAt', updatedAtName: 'gitUpdatedAt' }],
    '@nuxt/content',
  ]
}
```

Note that the module needs to be added **before** `@nuxt/content` because `@nuxt/content` needs to know about the hooks when loading.

pablomuro:
@dword-design I tried your solution on a Vercel deploy, but didn't work, logged the error "Error: spawn git ENOENT"
😢  I have to find another solution for the createdAt lost of data

dword-design:
@pablomuro Sounds like git is not installed on your deployment system. It needs to be there, otherwise the module cannot analyze the logs.

PR #548, opened 2020-10-08
fix: avoid registering `components/global` when not exists

### LINKED ISSUE #547: [theme-docs] avoid registering `components/global` when does not exists
### Version
@nuxt/content: 1.5.0
nuxt: 2.x

### Steps to reproduce

- yarn create nuxt-content-docs

### What is Expected?

Not to register `components/global` when does not exist and avoid warning 

> WARN Components directory not found: ~/components/global

pi0:
## Types of changes

## Description
Fixes #547 

## Checklist:

### DIFF
--- a/packages/theme-docs/src/index.js
+++ b/packages/theme-docs/src/index.js
@@ -47,13 +47,6 @@ const defaultConfig = {
       nuxt.options.content.dir = path.resolve(nuxt.options.rootDir, nuxt.options.content.dir || 'content')
       // Configure `static/ dir
       nuxt.options.dir.static = path.resolve(nuxt.options.rootDir, nuxt.options.dir.static || 'static')
-      // Configure `components/` dir
-      nuxt.hook('components:dirs', (dirs) => {
-        dirs.push({
-          path: path.resolve(nuxt.options.rootDir, 'components/global'),
-          global: true
-        })
-      })
       // Configure `tailwind.config.js` path
       nuxt.options.tailwindcss.configPath = nuxt.options.tailwindcss.configPath || path.resolve(nuxt.options.rootDir, 'tailwind.config.js')
       nuxt.options.tailwindcss.cssPath = nuxt.options.tailwindcss.cssPath || path.resolve(nuxt.options.rootDir, nuxt.options.dir.assets, 'css', 'tailwind.css')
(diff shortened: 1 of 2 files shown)


packages/theme-docs/src/index.js
```diff
@@ -47,13 +47,6 @@ const defaultConfig = {
       nuxt.options.content.dir = path.resolve(nuxt.options.rootDir, nuxt.options.content.dir || 'content')
       // Configure `static/ dir
       nuxt.options.dir.static = path.resolve(nuxt.options.rootDir, nuxt.options.dir.static || 'static')
-      // Configure `components/` dir
```
benjamincanac, line 50: @pi0 Why removing this?

The theme needs to load global components inside the theme itself and the ones from the `rootDir`.

pi0: You are right. Using `rootDir` was little bit misleading. I've updated it to `__dirname`

benjamincanac: @pi0 This has to be `nuxt.options.rootDir`, @nuxt/content is already loading the theme components, here `__dirname` also resolves to the theme so user components are not loaded anymore.

pi0: Hmm i not sure to unrestand. `rootDir` cannot point to two directories. `rootDir` should be same as `__dirname` (path to `node_modules/@nuxt/theme/docs/src`. User components are loaded by the other hook in main content module

benjamincanac: Since the `srcDir` is the theme, the main content module is loading the theme components, `rootDir` however loads the user components.

benjamincanac: You can try to create a component in the `docs/` for example, they won't be loaded with the `__dirname` change.

Atinux:
I will look at it too, it was in my todo list :)

I think we need to keep rootDir (user directory when using the docs theme).

Also I would like to watch the creation of the global/ directory in order to restart Nuxt.

On top of it, since we have the PR for overwriting components, we will have to wait for it to be merged since it will be two directories to watch

packages/theme-docs/src/index.js
pi0: ok you are right reverted :)

pi0:
> On top of it, since we have the PR for overwriting components, we will have to wait for it to be merged since it will be two directories to watch

I'm not sure it will we merged in near future because is a breaking change and also not related to this fix

Atinux:
You are right.

Can we watch the repository creation (if does not exists) to restart Nuxt: `nuxt.options.watch(join(nuxt.options.srcDir, 'components/global/'))`

What is the current behavior if you start Nuxt, then add a component to components/global/?

benjamincanac (approved):

benjamincanac (approved):

PR #574, opened 2020-10-16
fix(theme-docs): twitter preview when trailing slash on url

Atinux:
## Types of changes

## Description

When adding a trailing slash to `url`  in `content/setting.json`, the preview on Twitter was broken, this avoid this situation.

I also added a warning if the user forget to fill the url in `content/setting.json`.

## Checklist:

### DIFF
--- a/packages/theme-docs/src/layouts/default.vue
+++ b/packages/theme-docs/src/layouts/default.vue
@@ -28,6 +28,7 @@ export default {
   },
   head () {
     const i18nSeo = this.$nuxtI18nSeo()
+    const previewUrl = this.settings.url.replace(/\/$/, '') + '/preview.png'
 
     return {
       titleTemplate: (chunk) => {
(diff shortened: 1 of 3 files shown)


benjamincanac (approved):

packages/theme-docs/src/layouts/default.vue
```diff
@@ -28,6 +28,7 @@ export default {
   },
   head () {
     const i18nSeo = this.$nuxtI18nSeo()
+    const previewUrl = this.settings.url.replace(/\/$/, '') + '/preview.png'
```
pi0, line 31: This is implicitly making it mandatory to have `preview.png` in `static` dir without warning

Atinux: Indeed, it was already the case, I think we should improve this part: https://content.nuxtjs.org/themes/docs#static

pi0: Indeed but maybe we can also check it by module and show a warning or simply avoid adding previwUrl if file doesn't exists?

PR #598, opened 2020-10-23
fix(content): fix routing based on locale

### LINKED ISSUE #545: Docs links not including locale in path
### Version
Actual website

### Reproduction Link
https://content.nuxtjs.org/

### Steps to reproduce
Go to one of translated version of docs and click on an internal link (https://content.nuxtjs.org/fr/writing#markdown and first link in this section for example).

### What is Expected?
Should redirect to locale version of doc.

Redirects to default (english) version of docs, it omits the locale path part.

Should it be added in the links or the config is missing something ?

NozomuIkuta:
## Types of changes

## Description

This PR resolves #545.

As the reporter said, links inside markdown contents don't navigate to the pages in default locale.

I found that such links are correctly parsed as NuxtLink components, but the original paths are set to `to` prop, without converting them by `vue-i18n`'s `localePath()` function.

In the first commit, I fixed this bug by modyfing `packages/content/templates/nuxt-content.js`.
As far as I tested in the newly created `theme-docs` app, my fix works.

How I tested:

- Set up locale config, English and French, according to [the nuxt content official documentation](https://content.nuxtjs.org/themes/docs#nuxtconfigjs).
- Copy `content/en/` to `content/fr/` 
- Create links between `index.md` and `setup.md` in each locale
- Check whether navigation is done in the current locale
  - `index` to/from `setup`
  - `fr/index` to/from `fr/setup`
  -  switching locale in the footer

## Checklist:

Since I found i18n is not tested in the `@nuxt/content` package, I tried to add tests for that.
This is what I tried to do in the second commit.

Changes I made:

- Put contents of `fixture` directory into 2 sub directories, `default` and `i18n`.
  - `default` is where the original contents moved
- Fixed relative paths for `loadConfig()` function to work as before
- Set up `i18n` config in `fixture/i18n` directory
  - `i18n` module options are excerpted from `theme-docs`
- Run `yarn run test:packages:content`

As a result, I couldn't pass the test I created. The reasons was that `nuxt-i18n` is not transpiled in advance, so Jest couldn't parse the files, and said like "Do not use `import` outside of module". I found some related information and tried some, but couldn't resolved it. 😢 

Finally, I decided to create a draft PR because someone have already solved this kind of issue before.


NozomuIkuta:
Ah, I found I created the branch from other commit which is ahead of `nuxt/content/dev`.
I will fix it soon and push this branch again.

packages/content/templates/nuxt-content.js
```diff
@@ -109,7 +115,8 @@ export default {
       required: true
     }
   },
-  render (h, { data, props }) {
+  render (h, { data, props, parent }) {
+    const { localePath } = parent
```
NozomuIkuta, line 119: I'm not sure but will it broken if `NuxtContent` component is inside another functional component?

benjamincanac:
@NozomuIkuta Thanks for your PR, however this shouldn't be a fix in @nuxt/content because it does not depend on nuxt-i18n. This should be fixed in the theme-docs, I will take a look soon!

NozomuIkuta:
@benjamincanac 

Oh, I missed the point.

Thank you for your attention, and I wish this bug is fixed in near future because i18n is important for the web. 👍

PR #603, opened 2020-10-27
fix: #465 make dark theme accessible

### LINKED ISSUE #465: The (custom) dark theme primary color is not accessible
The (custom) dark theme primary color is not accessible.

### Version

all

### Steps to reproduce

Create a Tailwind CSS color palette with the `500` tint being accessible on a white background and inaccessible on a dark background.

### What is Expected?

In the dark mode, the primary color should be `400` instead of `500` to make it more accessible.

### What is happening?

The dark mode uses `500` as the primary color which causes many color palettes to be inaccessible.

mayashavin:
## Types of changes

## Description
- Resolves #465
- Allow user to define an additional dark primary color (if they don't wish to use the current primary generated colors)

## Checklist:

### DIFF
--- a/packages/theme-docs/src/tailwind.config.js
+++ b/packages/theme-docs/src/tailwind.config.js
@@ -24,7 +24,8 @@ module.exports = ({ docsOptions, nuxt }) => ({
         mono: ['DM Mono', ...defaultTheme.fontFamily.mono]
       },
       colors: {
-        primary: getColors(docsOptions.primaryColor || '#00CD81')
+        primary: getColors(docsOptions.primaryColor || '#00CD81'),
+        darkPrimary: getColors(docOptions.darkPrimaryColor || '#00CD81')
       },
       maxHeight: {
         '(screen-16)': 'calc(100vh - 4rem)'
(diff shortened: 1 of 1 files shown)


benjamincanac:
@mayashavin Thanks for the PR, however adding a `darkPrimary` color is not the solution as the `primary` color is used everywhere in the template. 

You can take a look at this file: https://github.com/nuxt/content/blob/dev/packages/theme-docs/src/components/app/AppNav.vue#L28, it demonstrate the use of the `text-primary` classes combined with the `dark` variant.

mayashavin:
@benjamincanac I understand how dark variant is used. Was thinking we may want users to have the options to provide a different theme color for dark mode, rather than simple rely on -400 variant. In any case, I can remove `darkPrimary` if you are ok with just use `-400` for dark mode only.

benjamincanac:
@mayashavin Do you think using the `400` as a base for `primary` will solve the accessibility issue?

mayashavin:
@benjamincanac For dark-mode, I think in most case it should. But again, it all depends on what is the base (primary-500) user defined in the config. I think we can go with 400 and see how it goes :)

benjamincanac:
@mayashavin Yeah going with the `400` as a default for the primary in dark mode is a good first step 🙂

benjamincanac:
@mayashavin This will be implemented in the next version of the theme.

PR #615, opened 2020-11-02
Add RTL support

amirmasoud:
This PR make the theme RTL compatible by adding the following block of code:

```js
...
  head: {
    bodyAttrs: {
      dir: 'rtl'
    }
  },
...
```
to the `nuxt.config.js` file.

## Types of changes

## Description
**Describe your changes in detail**
This PR adds `tailwindcss-dir` package to detect `ltr` and `RTL` version. Required TailwindCSS classes have been added so it won't break the already working LTR version.

**Why is this change required? What problem does it solve?**
An RTL version is not necessary at the moment for the documentation(s), as I have not seen the RTL version of documentations (neither website nor translations). But if it's in your internal team checklist for future features, this PR can be helpful.

I was going to use this theme for a Persian (RTL) website, and I thought this PR might be welcome. But there are some parts of it that might need more work and adjustments, such as Algolia search and tailwindcss-typography. As tailwindcss typography has an open issue for RTL support (https://github.com/tailwindlabs/tailwindcss-typography/issues/78)

## Checklist:

### DIFF
--- a/packages/theme-docs/package.json
+++ b/packages/theme-docs/package.json
@@ -28,7 +28,8 @@
     "vue-scrollactive": "^0.9.3"
   },
   "devDependencies": {
-    "nuxt": "^2.14.7"
+    "nuxt": "^2.14.7",
+    "tailwindcss-dir": "^4.0.0"
   },
   "types": "./src/index.d.ts"
 }
--- a/packages/theme-docs/src/assets/css/main.css
+++ b/packages/theme-docs/src/assets/css/main.css
@@ -85,5 +85,25 @@
   & pre[class*="language-"] {
     /* Make pre static so the relative goes to the parent (.nuxt-content-highlight) */
     @apply bg-gray-800 static;
+
+    direction: initial;
+  }
+}
+
+body[dir="rtl"] {
+  & .nuxt-content {
+    & h2, & h3 {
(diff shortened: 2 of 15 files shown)


benjamincanac:
@amirmasoud Thank you for the PR, it would be a nice add to the theme indeed. However I think we should we wait for the issue on `tailwindcss-typography` plugin to be fixed to have the full support.

amirmasoud:
@benjamincanac I see; I would try to submit a PR to `tailwindcss-typography` to add RTL support. I would get back to this PR after that. Thanks,

amirmasoud:
I don't think the Tailwind team has any plan to add RTL support to the core package for now. https://github.com/tailwindlabs/discuss/issues/64#issuecomment-532801714

Besides, there is no `direction` property by default in the new version of the tailwind. I don't think it's a good idea to put a time to add an RTL version for now. I would close this PR for now. Thanks.

ahangarha:
May you please share your thought on the best possible way to address this issue? Should we propose something to tailwindcss?

amirmasoud:
@ahangarha As far as I can tell, the majority of works fall into the tailwind typography package which is not part of nuxt/content obviously and it seems they have no plan to add RTL support for that.

Issue #658, opened 2020-11-22
feat(content): add a way to fetch surrounding contents by path instead of slug


NozomuIkuta:
### Is your feature request related to a problem? Please describe.

This feature request is based on [my comment](https://github.com/nuxt/content/issues/641#issuecomment-730258400) to #641. 

> @-Atinux
> 
> Sorry to bother you. 🙇
> 
> But, as far as I know, there is no explanation in docs that slugs are supposed to be unique in content directory, for fetch() with surround query to work as we expect.
> 
> Is it kind of common sense in web industry that slug is always unique?
> Actually, I think it is natural that multiple category has same slug, like @uta-mori 
sts-ryan-holton does.
> 
> Fetching documents, always based on path rather than slug, seems to let developers create files in more flexible way, especially when > their routing depends on file/directory structure, as Theme Docs does.
> 
> What do you think on this topic, or what is your original intention to use slug to fetch surrounding documents?
> 
> [Reproduction Link (Wrong Prev/Next Link) (CodeSandbox)](https://codesandbox.io/s/upbeat-turing-kl7x3)

_Note: I replaced `@` mention with `@-` to avoid notification to the people._

### Describe the solution you'd like

I would like Nuxt Content to support fetching surrounding documents by path instead of slug, so that we can define our website architecture more flexibly (e.g. making a web book, where each chapter has "introduction" page).

### Describe alternatives you've considered

Instead of adding a new query function, it might work to let the existing `surround()` function accept path as first argument.

```javascript
await $content().surround(slugOrPath, { before: 1, after: 1 }).fetch()
```

I'm not sure how to distinguish path from slug, though (maybe existence of `/`?).

### Additional context

I'm mainly talking about markdown content, so not sure about situation of other type contents.

PR #660, opened 2020-11-24
feat(content/theme-docs): support surround by path

NozomuIkuta:
## Types of changes

## Description

Resolves: #658. 

## Checklist:

### DIFF
--- a/docs/content/en/fetching.md
+++ b/docs/content/en/fetching.md
@@ -161,16 +161,16 @@ Check out [this snippet](/snippets#search) on how to implement search into your 
 
 </alert>
 
-### surround(slug, options)
+### surround(slugOrPath, options)
 
-- `slug`
+- `slugOrPath`
   - Type: `String`
   - `required`
 - `options`
   - Type: `Object`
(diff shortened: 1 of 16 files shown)


NozomuIkuta:
Oops, I've found that this PR includes breaking change...renaming argument... (i.e. `slug` to `slugOrPath`)

Should I revert the name to `slug` and let the method accept path string as well? Or should I create another method like `surroundByPath()`?

NozomuIkuta:
Naming may be an issue in terms of API body.
So, another alternative would be accepting both `slug` and `path` and checking which is passed.

Atinux:
Thank you for this great PR @NozomuIkuta 

I don't think renaming is really a breaking change since people use `$content` and not directly making API call to `/_content`  endpoint.

I am good with this PR, maybe @benjamincanac might see stuff that I don't see.

benjamincanac (approved):

benjamincanac:
@NozomuIkuta Thanks for this, I'm aware that it's been a bit complicated to implement the surround when working with sub-directories.

PR #671, opened 2020-12-03
feat: sort by score (most relevant first) when searching

danielsaul:
## Types of changes

## Description

I found when testing my search, that I was receiving strange results, with relevant ones mixed in with not so relevant ones. I spent a while digging, and found that the results are not being sorted at all.

When you perform a search, it is normally expected that the results will be sorted by the **most relevant** results first. 
The existing `search()` method does not do this, and it is not possible to use `sortBy()` to sort by the underlying full text search score from LokiDB.

LokiDB offers a method to `sortByScoring()` which must be used to ensure the most relevant results are returned.

https://github.com/LokiJS-Forge/LokiDB/blob/master/packages/loki/src/result_set.ts#L377
https://github.com/LokiJS-Forge/LokiDB/blob/master/packages/full-text-search/spec/generic/full_text_search.spec.ts#L200

Therefore `search()` will now call `sortByScoring()` within it by default, sorting by the search score descending, to provide the most relevant results first.

## Checklist:

### DIFF
--- a/packages/content/lib/query-builder.js
+++ b/packages/content/lib/query-builder.js
@@ -115,7 +115,8 @@ class QueryBuilder {
       }
     }
 
-    this.query = this.query.find({ $fts })
+    this.query = this.query.find({ $fts }).sortByScoring()
+
     return this
   }
 


danielsaul:
> Thank you for this PR @danielsaul
> 
> I think that `$content().search()` should actually call `.sortBySearchScoring()` by default to send back relevant results instead.

I did consider this, and is exactly what I did initially, and definitely happy to swap to that.

The reason I didn't was so that we could expose the direction parameter, and allow sorting by the score in ascending order. It's not really possible to add direction as an additional parameter to `search()`.

Having said that, not sure why you would ever want to sort by score ascending 🤷 so maybe not a concern.

I'll also double-check that sortByScoring won't have adverse affects on chaining a subsequent sortBy for example, but I can't imagine it would.

danielsaul:
> > Having said that, not sure why you would ever want to sort by score ascending 🤷 so maybe not a concern.
> 
> Exactly, sometimes the best option is no option 😄 And anyway they can always `.reverse()` the result if they want to.

Cool! Simplified and updated PR description 👍 
Don't think there's really any need for any documentation changes now.

Atinux (approved):

benjamincanac (approved):

Issue #457, opened 2020-09-05
Is there a way to import code from external file into a codeblock?


epigramengineer:
Is there a built in or recommended way to inline code from an external file in a codeblock for a markdown content page? I want to store a code snippet in a source file, but inline it in my markdown. Currently I'm able to load the code with the following:

```
<template>
  <div class="nuxt-content-highlight">
    <pre class="line-numbers"><code>{{code}}</code></pre>
  </div>
</template>

<script>
import raw from "@/assets/code/my_code_file";
export default {
  data {
    return { code: raw }
  }
}
</script>
```

which "works" and I could create a component, but it feels like I'm reinventing the wheel. I hardly found anything about external file loading in nuxt, much less for nuxt-content within markdown. Ideally I'd have a markdown file like the following:

<pre>
# My markdown file

```
{ include assets/code/my_code_file }
```
</pre>

And nuxt-content would do the heavy lifting for me.

kaka-ruto:
Interested on thi one too! Would be really cool if this worked.

Also you could specify a thing as range, because sometimes you do not want to import all lines.

So something like 
```
{ include assets/code/my_code_file.js[4-8] }
```

That would be interesting! Wonder if something like this exists

mathe42:
You can use the new hook created with #453 wich is included in the next release.

benjamincanac:
@epigramengineer Would be really nice to add this to the documentation if you guys can make it work! 🙂

danielroe:
This is one such example: https://github.com/nuxt/typescript/blob/master/docs/components/global/InjectCode.vue

danielroe:
@Atinux Happy to provide a PR if that would be helpful?

Atinux:
Yes please @danielroe 💚

PR #676, opened 2020-12-08
docs: add snippet

NozomuIkuta:
## Types of changes

## Description

Resolves: #457

I omitted the logic to set a range of lines to fetch, for snippet to be simple.
I believe people now are likely to come up with the way to achieve it. 😀 

```
{{ include assets/code/my_code_file.js[4-8] }}
```

### Questions

- Is my approach reasonable and compatible with Nuxt way?
- Do section titles (i.e. "Advanced" and "Dynamic Code Block") look fine?
- Should I implement the snippet as a built-in feature of Theme Docs?

### Help Wanted

- Would you help me write English sentences to make the section more informative? 🙏 

## Checklist:

I saw the snippet working in my local with Nuxt app scaffolded by `create-nuxt-content`.

### DIFF
--- a/docs/content/en/snippets.md
+++ b/docs/content/en/snippets.md
@@ -306,3 +306,86 @@ export default {
   }
 }
 ```
+
+## Advanced
+
+### Dynamic Code Block
+
+You can leverage `content:beforeParse` hook to manage source code separately from markdown files.
+
+```markdown[/content/en/index.md (original)]
+# Backslashes are for demonstration
+
+---
+title: Dynamic Code Block
+---
+
+\```
+source:./code-sample.js
+\```
+```
+
+```markdown[/content/en/code-sample.js]
+import Vue from 'vue'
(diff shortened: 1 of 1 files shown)


NozomuIkuta:
Ah...why do I always find some missing points after submitting PRs...? 😞 
(Maybe because I always create PRs after working during night, though)

Anyway, source code files are not watched in my approach, so I have to save the markdown file (even with nothing changed) every time I edit the source code files. I guess this is not acceptable DX.

But, in this time, I'm not sure whether there is such a way that the markdown file is loaded again... 🤔

NozomuIkuta:
OK, I found that Nuxt's `watch` option, so giving a pattern to source code files or enumerating them will resolve the reloading issue.
But, at the same time, we have to loose one of great DX of Nuxt module, that is, independency from Webpack (or bundling). 😕

PR #467, opened 2020-09-11
feat(content): add custom highlighter

jefrydco:
## Types of changes

- [ ]  Bug fix (a non-breaking change which fixes an issue)
- [x]  New feature (a non-breaking change which adds functionality)
- [ ]  Breaking change (fix or feature that would cause existing functionality to change)

## Description

This will add all of the features I described here #466  , with additional feature if we need to get the higlighter from promise-returned package/function.

```js
// nuxt.config.js
export default {
  content: {
    markdown: {
      async highlighter() {
          const customHighlighter = await getCustomHighlighter() 
          return (rawCode, lang) => customHighlighter(rawCode, lang)
      }
    }
  }
}

```

By adding this feature, now we can apply endless possibility to highlight our code. For instance, the gif below, I use `shiki-twoslash` by @orta combined with `v-tooltip` by @Akryum to add beautiful rich type information to `@nuxt/theme-docs`.

nuxt-content-custom-highlighter

Closes #466 

## Checklist:

- [x]  My change requires a change to the documentation.
- [x]  I have updated the documentation accordingly.
- [x]  I have added tests to cover my changes (if not applicable, please state why)

PR #680, opened 2020-12-13
fix(content): failed ci on windows and enhance file name detection regex

jefrydco:
## Types of changes

## Description

This PR do 2 things:
1. Fix failed CI build on windows.
Last time we added custom highlighter, the test case always make the CI failed on Windows. After long debugging, I found that the culprit was shiki twoslash always renders a custom link to TypeScript playground for each code block. I don't know why the custom link is generated a little bit random although the code isn't changed in Windows. It doesn't work that way in Linux or Windows, the link will always be same. That's why the CI isn't failed in Linux and Mac. So I decided to remove that link to make the test case worked.

To give more context @benjamincanac and I already discussed here #467 .

2. Enhance file name detection regex
Using previous method, this code won't parsed correctly:

```markdown
```typescript {}[something.ts] somemeta
console.log('Hello World')
```

By default we already parsed language, line highlight and file name. And actually we can still add our custom data/identifier there. In new custom highlighter syntax we can get that value through `node`:

```
export default {
  content: {
    highlighter(rawCode, lang, thematicBlock, astUtility) {
      const { node } = astUtility
      const { meta } = node 
      // meta here contains "somemeta" above
    }
  }
}
```

## Checklist:


Atinux (approved):
Beautiful work!

benjamincanac (approved):

PR #681, opened 2020-12-13
feat(theme-docs): add customizable slots

lupas:
This PR adds the possibility to add global components as "slots" to the left and right sidebar (nav + toc).

Possibile use cases: Sponsors, Ads, News feeds, Twitter feeds, etc.

## Types of changes

## Description
To make it easy, all we require is the following:
- Add any component to `components/global`
- Set `settings.slots.slotname = <componentName>`
  ```json
     "slots": {
      "navBarTop": "<COMPONENT_NAME>",
      "navBarBottom": "<COMPONENT_NAME>",
      "tocTop": "<COMPONENT_NAME>",
      "tocBottom": "<COMPONENT_NAME>"
    }
   ```

After that, the slot will be shown like so:

Screenshot 2020-12-13 at 21 25 11

## Checklist:
   --> imho not needed to add tests for this


lupas:
Created the following branch of the Firebase docs for quick testing:
https://github.com/nuxt-community/firebase-module/tree/nuxtContentDocsThemeSlotsExample/docs

Simply `cd docs && yarn && yarn dev` and then have a look. Ignore the rest of the repo, only `docs` folder is important.

Geminii:
Hi @lupas, 
A good idea to add content into some container. 
What about responsive position ? In mobile for example about the navbar ?

lupas:
Hey @Geminii 

The slots are also present in mobile view at the right positions:
Screenshot 2020-12-14 at 17 12 59
Screenshot 2020-12-14 at 17 13 48

Could of course make the top margins in navBarBottom a bit nicer, but imho already almost as expected.

If one wants different behaviour depending on responsiveness I would leave this to be handled inside the component that is loaded.

Or any other ideas?

Geminii:
About the margin, i think we can add it with a default configuration and maybe override with parameters ? 
Indeed the behavior about responsiveness could be handle directly inside the component to permit more flexibility 😀 

Great idea to purpose this feature 👌

lupas:
@Geminii 
Also the margins can be handled within the components, no? If someone wants additional margin, they can just add or substrract it to/from their component. I don't think we need to make the configuration more complex for that.

In the newest commit I fixed the default margins, made the slots simpler (by moving all the logic to an AppSlot component) and also added 4 more slots:
Screenshot 2020-12-17 at 19 25 24

There might be an even simpler way to implement this by e.g. just scanning the folder `global/slots` for the existence of properly named components. But I think the way it is now is better so people can e.g. use the same components in two slots.

@benjamincanac @Atinux What do you think?

Atinux:
I do believe it is a good idea @lupas 

I am thinking about the best implementation for it 🤔

Geminii:
Indeed there is a little bit too many slot now. This need to be discuss with the team about new creation of doc theme. 
The idea behind this is to permit more flexibility about the creation of a beautiful custom documentation ❤️

lupas:
@Geminii 
Yep, we can always remove some of the slots again if we think it's too much, I agree.
The way it is implemented needs just once line of code to create a new slot, so it's not cluttering the code too much.

@Atinux 
I was thinking about other options to implement that too (e.g. by wrapping all in a template and using v-slots, but that would need an entire rewrite of the structure - so I opted for this option).
If you come up with a better solution to implement this let me know :)

PR #692, opened 2020-12-21
docs: add missing description

NozomuIkuta:
## Types of changes

## Description

There is no explanation about `database` argument of `content:file:beforeInsert` hook, which was introduced by #442.
This PR is to add description of it.

## Checklist:

I checked UI in my local by running `yarn run dev` in `docs` directory.

### DIFF
--- a/docs/content/en/advanced.md
+++ b/docs/content/en/advanced.md
@@ -102,6 +102,10 @@ Arguments:
   - Type: `Object`
   - Properties:
     - See [writing content](/writing)
+- `database`
+  - Type: `Object`
+  - Properties:
+    - See [type definition](https://github.com/nuxt/content/blob/master/packages/content/types/database.d.ts)
 
 **Example**
 


Geminii:
Hi @NozomuIkuta 👋 
Thanks to add a section about Database. Instead of add a link to the Database type, maybe we can create a page dedicated to the database explanation no ? What do you think about it ?

NozomuIkuta:
@Geminii 

Hi, thank you for reviewing my PR. 😄 

> Instead of add a link to the Database type, maybe we can create a page dedicated to the database explanation no ? What do you think about it ?

Yes, I agree with you basically. Actually, I directed the link to type definition simply because there was no dedicated page in the docs.

By the way, how about adding a dedicated section to _Advanced_ page instead?
I think, in most of use cases, people would not have to know what Nuxt Content's database looks like and how it works.

In my humble opinion, the points below would be worth mentioning:

- Database owns parsers
  - This point is already implied by the example in [the section of `content:file:beforeInsert`](https://content.nuxtjs.org/advanced#contentfilebeforeinsert).
  - This content would provide further information about `extendParser` option
- Database stores files in an array
  - This point is already implied by existence of `sort()` method
  - Additionally, I also implied this point in #660 (see changes on `docs/content/en/snippets.md`)
- Lifecycle of a file
  - (walk) ➡️ parse ➡️ (update) ➡️ insert ➡️ (remove)
  - These may not be related straightforward, though 🤔

Atinux:
Thank you @NozomuIkuta :)

I don't think we need to add a section about the database since I am planning to find an alternative of LokiJS and Database might be different for the version 2 of nuxt content.

PR #699, opened 2020-12-29
fix(content): robustify filter regex for $content( path, { deep: true } ) 

flowrakis:
The path in $content(path) is now regex-escaped befor being fed to LokiJs.
This Change escapes those characters with .https://stackoverflow.com/questions/3561493/is-there-a-regexp-escape-function-in-javascript/3561711

## Types of changes

## Checklist:

### DIFF
--- a/packages/content/lib/database.js
+++ b/packages/content/lib/database.js
@@ -46,7 +46,7 @@ class Database extends Hookable {
   query (path, { deep = false, text = false } = {}) {
     const isDir = !path || !!this.dirs.find(dir => dir === path)
     // Look for dir or path
-    const query = isDir ? { dir: deep ? { $regex: new RegExp(`^${path}`) } : path } : { path }
+    const query = isDir ? { dir: deep ? { $regex: new RegExp(`^${path.replace( /[.*+?^${}()|[\]\\]/g, '\\$&') }`) } : path } : { path }
     // Postprocess to get only first result (findOne)
     const postprocess = isDir ? [] : [data => data[0]]
 
--- a/packages/content/lib/database.js
+++ b/packages/content/lib/database.js
@@ -46,7 +46,7 @@ class Database extends Hookable {
   query (path, { deep = false, text = false } = {}) {
     const isDir = !path || !!this.dirs.find(dir => dir === path)
     // Look for dir or path
-    const query = isDir ? { dir: deep ? { $regex: new RegExp(`^${path.replace( /[.*+?^${}()|[\]\\]/g, '\\$&') }`) } : path } : { path }
+    const query = isDir ? { dir: deep ? { $regex: new RegExp(`^${path.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') }`) } : path } : { path }
     // Postprocess to get only first result (findOne)
     const postprocess = isDir ? [] : [data => data[0]]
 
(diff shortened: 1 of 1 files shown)


flowrakis:
Hey @Atinux, of course. It would be my pleasure to provide a testcase (especially since this PR would remove a show-stopper for us.) But I must admit, I am not shure where and how. Could I ask you to point me that way? regards, Emanuel.

packages/content/lib/database.js
```diff
@@ -46,7 +46,7 @@ class Database extends Hookable {
   query (path, { deep = false, text = false } = {}) {
     const isDir = !path || !!this.dirs.find(dir => dir === path)
     // Look for dir or path
-    const query = isDir ? { dir: deep ? { $regex: new RegExp(`^${path}`) } : path } : { path }
+    const query = isDir ? { dir: deep ? { $regex: new RegExp(`^${path.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')}`) } : path } : { path }
```
NozomuIkuta, line 49: JFYI: 

Lodash has such function that escapes a regular expression: `escapeRegExp`.
This would be more [well-tested](https://github.com/lodash/lodash/blob/2f79053d7bc7c9c9561a30dda202b3dcd2b72b90/test/escapeRegExp.test.js) by the team.
Note that I don't mean to import Lodash, just quoting (maybe with a comment for their MIT license).

References:

- [Definition of the expression to escape](https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L152)
- [Definition of the function](https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L14273-L14278)

flowrakis: @NozomuIkuta Thank you for your feedback. I agree, importing lodash would be overkill. Furthermore the regex escaping needed for matching filenames is not that complex, because all it needs to do is to escape the regex reserved characters.

Could you maybe tell me how and where I would best provide a test in the code-base, so that this PR could be accepted?

NozomuIkuta: @flowrakis 

You have modified how query is set and handled.
As far as I know, this kind of logic is tested in the following 2 files:

- [packages/content/test/module.test.js](https://github.com/nuxt/content/blob/main/packages/content/test/module.test.js) to test fetching logic via [API endpoints](https://content.nuxtjs.org/advanced#api-endpoint)
- [packages/content/test/server.test.js](https://github.com/nuxt/content/blob/main/packages/content/test/server.test.js) to test fetching logic via [`$content` function](https://content.nuxtjs.org/fetching)

I guess new tests would look like ones that fetch documents with `path` parameter for API and `$content()` including characters that must be escaped, and that check documents returned are expected.

Since the escaping function is "embedded", we can't test it alone (i.e. unit test) 
If we want or need, we will have to define it as a utility function in [packages/content/lib/utils.js](https://github.com/nuxt/content/blob/main/packages/content/lib/utils.js) and create a new test file for utilities. But, IMHO, it is not needed this time.

---

Note that I'm not a core team member, but only a contributor, so what I say is based on my personal experience. But, at least I can say that test files of Nuxt Content are stored in `test` directory.

NozomuIkuta: @flowrakis 

Is there anything that I could help?

Tahul:
Hello;

As this PR is pretty old now and we are releasing @nuxt/content v2, I'll be closing it.

Regex in queries are supported in v2 since this commit: https://github.com/nuxt/content/commit/df73c480196bf0e072ba0883deab388e81d95e45

Feel free to re-open an issue or this PR if you want to continue the work!

If regex queries does not suit your use case, could you please explain it?

PR #715, opened 2021-01-15
feat(theme-docs): adding styles to scrollbar

lupas:
The scrollbars of nuxt-content docs-theme currently looks like this:

Screenshot 2021-01-15 at 12 08 59

This is not really nice, especially when the side navigations get higher than the screen size.

This PR is an idea of how to nicely style the sidebars, so the overall experience of the docs-theme looks more professional. 

⚠️ I'm not the best styler, so I'm sure this could be further improved, maybe the primary color is not the best pick for the scrollbar? Feel free to change it to something even nicer looking.

## Types of changes

## Description

This PR makes the scrollbar look like this on Safari, Chrome and Firefox: 
Screenshot 2021-01-15 at 12 00 29
Screenshot 2021-01-15 at 12 01 22
## Checklist:
- [NOT NEEDED] My change requires a change to the documentation.
- [NOT NEEDED] I have updated the documentation accordingly.
- [NOT NEEDED] I have added tests to cover my changes (if not applicable, please state why)

### DIFF
--- a/packages/theme-docs/src/assets/css/main.css
+++ b/packages/theme-docs/src/assets/css/main.css
@@ -11,6 +11,29 @@
   display: block;
 }
 
+/* Style Scrollbar: Works on Firefox */
+* {
+  scrollbar-width: thin;
+  scrollbar-color: var(--color-primary-500) transparent;
+}
+
+/* Style Scrollbar: Works on Chrome, Edge, and Safari */
+*::-webkit-scrollbar {
+  width: 4px;
+}
+
+*::-webkit-scrollbar-track {
+    background: none;
+}
+  
+*::-webkit-scrollbar-thumb {
+    background-color: var(--color-primary-500);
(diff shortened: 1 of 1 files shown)


Atinux:
Thank you for the PR @lupas 

We will take a look at it this week, we planned a big change for the docs theme :)

lupas:
@Atinux 
Nice - looking forward to the big change <:o)

benjamincanac:
@lupas Thank you for this, I really like it! Do you think it could be possible to only show the bar when scrolling to have the macOS behaviour (when selected)?

Screenshot 2021-01-22 at 15 26 33

lupas:
@benjamincanac 
Sorry for the late response. I had a quick look at it but it seems like it would only be possible via Javascript - not sure if that's worth it. I'll have a closer look at it again after docus is out :)

PR #733, opened 2021-01-26
feat(content): Optional search parameter

Billybobbonnet:
## Types of changes

## Description

Passing an empty string  as query value will avoid triggering a search to allow optional chaining such as this:

```js
      this.total = (await this.$content('people')
          .search('')
          .only([])
          .fetch()).length
```
Docs have been updated in FR & EN. JA & RU update necessary.

## Checklist:

### DIFF
--- a/docs/content/en/fetching.md
+++ b/docs/content/en/fetching.md
@@ -139,7 +139,7 @@ const articles = await this.$content('articles').skip(5).limit(5).fetch()
 ### search(field, value)
 
 - `field`
-  - Type: `String` | `Boolean`
+  - Type: `String`
   - `required`
 - `value`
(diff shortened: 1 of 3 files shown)


Billybobbonnet:
I went for an explicit `false` parameter but maybe it would be more elegant and concise to skip the search if there is no parameter or an empty string. It can be fixed, let me know.

Atinux:
What about an empty string @Billybobbonnet, doesn't it work with it?

Billybobbonnet:
> What about an empty string @Billybobbonnet, doesn't it work with it?

In the current version, nope.
image

I updated my PR to fix this.

Atinux (approved):

Issue #767, opened 2021-02-10
Content module is blocking startup for parsing


pi0:
Modules can return a promise or be async in order to prepare nuxt context but not for other runtime operations since it makes `nuxt dev` stuck until initial parsing is done. (nuxtjs.org for instances takes 9,2 seconds on a corei7 CPU to parse and during this time we are blocking build)

I think we can move [database.init](https://github.com/nuxt/content/blob/2037ee87edc39018646f12f2d02ece1d2f80c7c1/packages/content/lib/index.js#L112) to a parallel promise instead of awaiting in main body. (not sure if any internals need to be updated to ensure init is done or not)

Also would be nice if in future versions doing parse off-thread (using node workers) to avoid blocking even-loop.

ManUtopiK:
Sometimes I got this issue. I solve it by deleting the folder `node_modules/.cache`

Atinux:
This is not the case with content v2 anymore. For v1 I don't think we will solve that issue (lack of time).

Issue #694, opened 2020-12-24
Parse files in parallel of Nuxt build


Atinux:
### Is your feature request related to a problem? Please describe.

On big website with a lot of content, starting @nuxt/content can be slow, and on top of this, it will block the webpack build to start.

### Describe the solution you'd like

Nuxt content should start parsing files without blocking and only pause the Nuxt `build:done` hook until the parsing of files done so the first page served will have the correct data fetched.

Atinux:
Closing since implemented in version 2 coming.

PR #772, opened 2021-02-11
ssr built app uses serialized content database

kwiatkk1:
## Types of changes

## Description

When you have a large number of content files (100+), app startup slows down a lot. It hurts in a development mode (see #694 and #767), but also when you deploy a built app onto a production server.

My proposal is to:
- during build, if target is `server`: to serialize content's db into a file
- during startup, if not in `dev` mode: do not walk through content's dir and parse found files, but to load content from a file generated during build

## Checklist:


kwiatkk1:
It's in a draft, because:
- it breaks existing tests
- it hasn't got its own tests

But I hope it shows the general idea. I hope you can look at it and say is it something you want to have in the product?
If so, I'll refine this pull-request.

pi0:
Hi @kwiatkk1 thanks for PR! I think it would be really good idea to use hard-cache to reduce development startup time. I would suggest moving logic `load` logic into Database.init and use an option to enable hard-cache by constructor or `init` (maybe `cache: true`). Also i think we should use `generate:done` hook since with full-static smart rebuild, we only generate content second time (which needs fresh parsed data)

kwiatkk1:
Hi @pi0, I made some changes: 
- logic hidden behind the `useCache` flag,
- tests,
- sketch of documentation.

This feature doesn't apply to developer mode or the full static generate -- production build only -- I have no idea how to make it work with content reloading :/

Issue #292, opened 2020-07-25
Not compatible with express serverMiddleware


danielroe:
> When a custom serverMiddleware is added that exports an express instance, all `@nuxt/content` client-side requests hang rather than resolving.

### Version
@nuxt/content: v1.4.1
nuxt: v2.13.3

### Reproduction link

https://codesandbox.io/s/nuxtcontent-demo-x11rc

### Steps to reproduce

1. Create serverMiddleware that exports an express instance that uses json middleware, for example:
   ```js
   import express from 'express'
   const app = express()
   // Note that without following line this bug is not reproduced
   app.use(express.json())
   app.get('/api/test', (req, res) => res.end())
   export default app
   ```
2. Add it to `nuxt.config.js`:
   ```js
   serverMiddleware: [require.resolve("./api/test.js")],
   ```

### What is expected?

I expect adding additional endpoints as above should not interfere with the endpoint `@nuxt/content` adds.

### What is actually happening?

All `@nuxt/content` client-side requests hang rather than resolving.

mawangod:
same bug

danielroe:
I have just refreshed the reproduction accessible above https://codesandbox.io/s/nuxtcontent-demo-x11rc - it is still giving the same issue.

Below is a video of the behaviour:
![Screen Recording 2020-11-05 at 11 23 57](https://user-images.githubusercontent.com/28706372/98235643-023cf880-1f5a-11eb-8854-223ff797cd76.gif)

Atinux:
My bad, I thought the serverMiddleware was not working, I see the issue now. Will investigate.

NozomuIkuta:
@Atinux @danielroe 

I've looked into this issue to have a fun on weekend. 

If I understand it correctly, the root cause is that `express.json()` (i.e. [body-parser](https://github.com/expressjs/body-parser) package, or [raw-body](https://github.com/stream-utils/raw-body) package more accurately) consumes all the stream before the `req` is passed to Nuxt Content's middleware.

So, the event listeners below are never called, and consequently the promise remains pending forever:

https://github.com/nuxt/content/blob/ebaedede5935f597fe325322ef57f09f6222e1ff/packages/content/lib/middleware.js#L19-L28

This promise stuck can be avoided by adding a condition to if-block, because body-parser sets `req._body` as `true` when it populates `req.body`:

```javascript
if (['POST', 'PUT'].includes(req.method) && !req._body) { /* ... */ }
```

[...]

PR #790, opened 2021-02-22
fix(content): handle readable ended request

NozomuIkuta:
## Types of changes

## Description

Resolves: #292

## Checklist:

~~I added a test to check default options.~~
~~If there is a way to check page transition with server middleware, please let me known so that I can add such test.~~
I removed `reqBodyKey` according to a review comment that `req.body` is a standard in Node.js community.

By the way, test was failed due to highlighter snapshot tests.
A color code was not matched in HTML.
I guess CI will fail too.

### DIFF
--- a/docs/content/en/configuration.md
+++ b/docs/content/en/configuration.md
@@ -524,7 +524,17 @@ Your component should implement the following:
 
 You should be aware that you get the full markdown file content so this includes the front-matter. You can use `gray-matter` to split and join the markdown and the front-matter.
 
+### `reqBodyKey`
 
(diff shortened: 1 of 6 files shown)


Atinux:
Thanks for the fix @NozomuIkuta 

I think we don't need `reqBodyKey` since `req.body` is a standard in Node :)

Atinux (approved):

pi0 (approved):

pi0:
Thanks! 

Note for future: We shall use [h3.useBody](https://github.com/nuxt-contrib/h3) utility later on

Issue #44, opened 2020-05-25
Support for i18n


uporot1k:
Hello, thanks for such a wonderful module. I have a question, is there any possibility to use i18n together with your module?

juliomrqz:
If your using [nuxt-i18n](https://github.com/nuxt-community/nuxt-i18n), you can save you files in different directories based on your supported languages and fetch them like this.

```js
export default {
  async asyncData({ app, params, $content }) {
    const { slug } = params;
    const posts = await $content("blog", app.i18n.locale, slug).fetch();

    return { posts };
  },
};
```

uporot1k:
There is my code 
`async fetch() {
    this.submenu = await this.$content('submenu', this.$i18n.locale, 'index').fetch().catch();
  },`

but when i switch page, content wouldn't update(

Atinux:
@juliomrqz answer is the one I recommend, and also the one we use in the content documentation: https://github.com/nuxt/content/blob/master/docs/pages/_slug.vue#L31

uporot1k:
> @juliomrqz answer is the one I recommend, and also the one we use in the content documentation: https://github.com/nuxt/content/blob/master/docs/pages/_slug.vue#L31

But what if i want use this module not on page component. I have nav which i want place in .md file and get it in fetch function.

PR #797, opened 2021-02-25
docs(docs): add l10n and i18n snippet

Tommytrg:
## Types of changes

## Description

I spent some time looking to use Nuxt content with @nuxt/i18n and found the solution in #44. I think having that response in the documentation would have saved me time.

## Checklist:

### DIFF
--- a/docs/content/en/integrations.md
+++ b/docs/content/en/integrations.md
@@ -284,3 +284,22 @@ We recommend to take a look at this tutorial made by [Pascal Cauhépé](https://
 👉 &nbsp;https://nuxt-content-and-forestry.netlify.app
 
 </alert>
+
+## Localization & Internationalization
+
+You can use Nuxt Content with [nuxt-i18n](https://github.com/nuxt-community/i18n-module) to support multiple languages.
+
+** Example ** 
+
+```
+export default {
+  async asyncData({ app, params, $content }) {
+    const { slug } = params
+    const posts = await $content(app.i18n.locale, "blog", slug).fetch()
+
+    return { posts }
+  },
+}
+```
+
+That code loads the content from the file located in the following path: `<locale>/blog.md`.


docs/content/en/integrations.md
```diff
@@ -284,3 +284,22 @@ We recommend to take a look at this tutorial made by [Pascal Cauhépé](https://
 👉 &nbsp;https://nuxt-content-and-forestry.netlify.app
 
 </alert>
+
+## @nuxtjs/i18n
+
+You can use Nuxt Content with [nuxt-i18n](https://github.com/nuxt-community/i18n-module).
```
Soviut, line 290: I realize it's inconsistent in this document, but it would be nice to include just a bit more of an explanation of what the i18n module is for.

```suggestion
You can use Nuxt Content with [nuxt-i18n](https://github.com/nuxt-community/i18n-module) to support multiple languages.
```

docs/content/en/integrations.md
```diff
@@ -284,3 +284,22 @@ We recommend to take a look at this tutorial made by [Pascal Cauhépé](https://
 👉 &nbsp;https://nuxt-content-and-forestry.netlify.app
 
 </alert>
+
+## @nuxtjs/i18n
```
pi0, line 288: Package name is (yet) not `@nuxtjs/i18n`). Maybe we can simply use `Localization` for title?

Tommytrg: Sounds good to me. Do you prefer `Localization`, `Internationalization` or simply `nuxt-i18n` for the title?

Soviut: While normally I'd agree that `Localization` is the name to go with, the module referenced directly after has `i18n` in the name. Perhaps the title could be both `Localization & Internationalization`?

Tommytrg:
What are the next steps for this?

NozomuIkuta (commented):
@Tommytrg 

Hi, 

I found a phrase below in Community/Snippet page, so I think it's nice to add a link from here to the content of your PR, isn't it (with modification if needed) ?

> Don't forget to prefix your calls with the current locale if you're using nuxt-i18n.

With this point changed, I think we can merge this PR. 👍

Tahul:
I'm merging this, will be useful for upcoming merging of docs (v1 / v2 on same website).

PR #820, opened 2021-03-10
feat: add yarn v2 (berry) support

mathe42:
## Types of changes

## Description

This PR adds yarn v2 support. For this the json5 dependencie has to be added directly in package.json

## Checklist:


mathe42:
As json5 is required by some code with yarn v2 @nuxt/content has to declare it in the dependencies.

This is a single line change to the package.json

Just wondering why is yarn.lock not in git and not in .gitignore? One of this should be true!?

NozomuIkuta (commented):
@mathe42 

Would you share some references so that we can know better why `json5` package is needed?
Once it's verified, I think I can merge this PR.

By the way, `yarn.lock` was generated again after you had created this PR.
So it would be safer if you merge the main branch to your branch and update `yarn.lock` file as needed.

mathe42:
JSON5 is imported here: https://github.com/nuxt/content/blob/main/packages/content/lib/database.js#L5 so for yarn v2 (berry) it would be optimal to add it to dependencies. (You can have multiple versions of a package with yarn v2 so yarn doesn't know wich to use. (There are workarounds but I think adding the dependencie would be a good idea).

NozomuIkuta (approved):
@mathe42 

Thank you for quick response!

I see, and yeah, JSON5 package definitely should be added as a dependency if it's directly used by the package, regardless of Yarn's requirement. 🙌

Issue #816, opened 2021-03-06
Incompatible type definition of fetch()?


Yz4230:
[Helped by DeepL]

Thank you for developing such a great module.
The type definitions for the creation and update dates did not work correctly, so I would like to ask if this is the intended behavior.

Here is the type definition of `QueryBuilder.fetch()` in [query-builder.d.ts](https://github.com/nuxt/content/blob/939caf36c547a6b4af6e303c52ed5989a5dea2f0/packages/content/types/query-builder.d.ts#L10)
```typescript
export class QueryBuilder {
// ...L80~L81
  fetch(): Promise<IContentDocument | IContentDocument[]>;
  fetch<T>(): Promise<(T & IContentDocument) | (T & IContentDocument)[]>;
}
```

And the return type of `fetch()` implementation and `IContentDocument` 's definition are ...
[query-builder.js](https://github.com/nuxt/content/blob/939caf36c547a6b4af6e303c52ed5989a5dea2f0/packages/content/lib/query-builder.js#L202)
```typescript
class QueryBuilder {
// ...L202~L246
  async fetch() {
  //...L245
    return JSON.parse(JSON.stringify(data))
  }
}
```

[content.d.ts](https://github.com/nuxt/content/blob/939caf36c547a6b4af6e303c52ed5989a5dea2f0/packages/content/types/content.d.ts#L6)
```typescript
//...L6
export interface IContentDocument extends Record<string, any> {
  dir: string;
  path: string;
  extension: '.md' | '.json' | '.yaml' | '.xml' | '.csv' | string;
  slug: string;
  createdAt: Date;
  updatedAt: Date;
}
```

In the `return` part of the `fetch()`, `JSON.parse(JSON.stringify(data))` is used to clone `data`. However, the actual return type from this implementation is `RealIContentDocument | RealIContentDocument[]`

```typescript
interface RealIContentDocument {
  dir: string;
  path: string;
  extension: '.md' | '.json' | '.yaml' | '.xml' | '.csv' | string;
  slug: string;
  createdAt: string; // This field cannot be `Date`. The value of this field seems to be a string of ISO time.
  updatedAt: string; // This field is too!
}
```

This is my first issue and I am sure I have some inadequacies, but I would appreciate it if you could help me.
Thank you.

Edit: This behavior seems to come from Nuxt. I continue investigating.

PR #825, opened 2021-03-14
Update type definition of QueryBuilder.fetch() (#816)

Yz4230:
## Types of changes

- [x] **NEW!** Update type definition

## Description

Just fix `QueryBuilder.fetch()` return type definition.

## Checklist:

### DIFF
--- a/packages/content/types/query-builder.d.ts
+++ b/packages/content/types/query-builder.d.ts
@@ -1,10 +1,23 @@
-import type { IContentDocument } from './content';
 interface QueryBuilderOptions {
   query: any;
   path: string;
   init: any;
   text: any;
   postprocess?: any[];
+}
+
+interface FetchReturn extends Record<string, any> {
+  dir: string;
+  path: string;
+  extension: '.md' | '.json' | '.yaml' | '.xml' | '.csv' | string;
+  slug: string;
+  createdAt: string;
+  updatedAt: string;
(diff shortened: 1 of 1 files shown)


packages/content/types/query-builder.d.ts
```diff
@@ -7,6 +6,20 @@ interface QueryBuilderOptions {
...
+  path: string;
+  extension: '.md' | '.json' | '.yaml' | '.xml' | '.csv' | string;
+  slug: string;
+  createdAt: string;
+  updatedAt: string;
+  toc?: {
+    id: string;
+    depth: number;
+    text: string;
+  }[];
```
danielroe, line 20: This should probably be added to `IContentDocument` instead of duplicating it here.

Yz4230: I will introduce `IContentDocumentBase` to avoid code duplication.
And add missing `body` field too.

packages/content/types/query-builder.d.ts
```diff
@@ -1,4 +1,5 @@
-import type { IContentDocument } from './content';
+import { IContentDocumentBase } from "./content";
```
danielroe, line 1:
```suggestion
import type { IContentDocumentBase } from "./content";
```

danielroe (approved):
Looks good to me - thank you ❤️

PR #894, opened 2021-05-22
fix(content): process dynamic component

### LINKED ISSUE #856: It is possible to pass through a component via frontmatter?
have a markdown file with YAML frontmatter. In the frontmatter, I want to refer to a component such as:

```
---
title: Hello world
component: <component-a></component-a>
---
Lorem ipsum...
```

I want to refer to this in a page file via {{page.component}}. Is there a way to have vue recognize this? Right now, it just spits out <component-a></component-a>, instead of the actual component.

NozomuIkuta:
## Types of changes

## Description

Resolves: #856 

I tried to resolve #856 by using Vue's `component` built-in component, by which we can have a control on what is rendered in runtime.
As a result, I failed and found the following points (if I'm correct):

- While `component` component is explained as a built-in component in term of API, it's actually not registered by `Vue.component()` like normal components.
- When we use `component` in Vue template, it is parsed like:
  - `tag` is value of `is` prop 
  - `data.attr.tag` is set as `"component"`.
- On the other hand, when we use `component` in markdown, the `node` object parsed is like:
  - `tag` is `"component"`
  - `props.is` is value of `is` prop
- As a result, Vue tries to search a component whose name is `"component"` and warns it's not registered.

This PR is to make consistent schema of parsed data, so that Vue can handle it correctly that `component` used in markdown.

...By the way, I have just come up with another approach while writing this message.
Should I have modified files in `parsers/markdown` directory...? 😨 

## Checklist:

It may be worth adding a snippet of dynamic component in markdown. Should I?

I have not yet added unit tests, but do some smoke tests with Nuxt app in my local: 

- Check if bare `<component :is="...">` in markdown is rendered correctly
- Check if `<component :is="...">` with slot content in markdown is rendered correctly

### DIFF
--- a/packages/content/templates/nuxt-content.js
+++ b/packages/content/templates/nuxt-content.js
@@ -5,6 +5,8 @@ const rootKeys = ['class-name', 'class', 'style']
 const rxOn = /^@|^v-on:/
 const rxBind = /^:|^v-bind:/
 const rxModel = /^v-model/
+const rxIs = /^is|^:is|^v-bind:is/
+const DYMAMIC_COMOPNENT = 'component'
 const nativeInputs = ['select', 'textarea', 'input']
 
 function evalInContext (code, context) {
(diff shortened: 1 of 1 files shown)


packages/content/templates/nuxt-content.js
```diff
@@ -79,6 +81,24 @@ function slotsToData (node, h, doc) {
...
+  const key = Object.keys(props).find((key) => rxIs.test(key))
+  const value = props[key]
+
+  if (value) {
+    node.tag = value in doc ? doc[value] : evalInContext(value, doc)
+    node.props.tag = DYMAMIC_COMOPNENT
+    node.children = node.children.filter((child) => {
+      const _child = processNode(child, h, doc)
+      return !(typeof _child === 'string' && /^\s*$/.test(_child))
+    })
```
NozomuIkuta, line 95: If I used `component` component in Vue template, the parsed node's `children` doesn't include nodes whose type is `"text"` and value is just like `\n`.

packages/content/templates/nuxt-content.js
```diff
@@ -79,6 +81,24 @@ function slotsToData (node, h, doc) {
...
+    delete node.props[key]
```
NozomuIkuta, line 96: If I used component component in Vue template, the parsed node doesn't have `is` prop.

packages/content/templates/nuxt-content.js
```diff
@@ -79,6 +81,24 @@ function slotsToData (node, h, doc) {
...
+  } else { /* Let Vue warn if :is prop is absent */ }
```
NozomuIkuta, line 97: Vue will warn like:

> [Vue warn]: Unknown custom element: <component> - did you register the component correctly? For recursive components, make sure to provide the "name" option.

Tahul:
Hey @NozomuIkuta ; on your own suggestion I am closing this! 😄

Let's keep in track in #856 if it is still relevant to `v2`. 🙂

Issue #904, opened 2021-06-04
CSS classes in Markdown component don't merge with component's root class


ManasMadrecha:
### Version
@nuxt/content: v1.14.0
nuxt: v2.15.5

### Steps to reproduce
1. Create a component `abc.vue`.
```html
<template>
  <div class="level1">
   <div class="level2">
    <slot></slot>
   </div>
  </div>
</template>
```

2. Use it in markdown

```md
<abc class="tw-bg-pink-100">
</abc>
```

### What is Expected?
1. `<abc>` component's root should have two classes, i.e. `level1` and `tw-bg-pink-100`.

### What is actually happening?
1. `<abc>` component only gets the class applied in markdown, i.e. `tw-bg-pink-100`.
2. `level1` class is simply removed.

NozomuIkuta:
Duplicate of #831?

PR #905, opened 2021-06-04
fix(content): merge css classes

### LINKED ISSUE #831: Vue component classes overriden by end-user specified classes in Markdown page
### Version
@nuxt/content: 
nuxt: 

### Reproduction Link

https://codesandbox.io/s/romantic-fermat-c3t5g
https://github.com/marceau-maubert/nuxt-content-component-class-override-bug

### Steps to reproduce

Create a Vue component that has CSS classes set on its root element, and register it globally in Vue so the Nuxt Content component can make use of it in our page.

Try to add additional CSS classes to the component when using it in a Markdown page via the `class` attribute.

### What is Expected?

The CSS classes should merge with the additional ones specified in the Markdown page.

The CSS classes will get overridden with the ones specified in the Markdown page.

It might be safe to assume something similar might happen with the `style` attribute, but I haven't tried it.

NozomuIkuta:
## Types of changes

## Description

Resolves #831 (thus, #904)

## Checklist:


packages/content/templates/nuxt-content.js
```diff
@@ -5,6 +5,7 @@ const rootKeys = ['class-name', 'class', 'style']
 const rxOn = /^@|^v-on:/
 const rxBind = /^:|^v-bind:/
 const rxModel = /^v-model/
+const rxClass = /^(?::|v-bind:)*class/
```
NozomuIkuta, line 8: These 3 cases will throw an error due to limitation of Nuxt Content.

```html
<my-component v-bind:class="{ 'my-class': true }">
<my-component v-bind:class="[ 'my-class' ]">
<my-component v-bind:class="'my-class'">
```

Should we only detect just `"class"` attribute?

packages/content/templates/nuxt-content.js
```diff
@@ -37,6 +38,8 @@ function propsToData (node, doc) {
       obj[field] = evalInContext(value, doc)
       data.on = data.on || {}
       data.on[event] = e => doc[value] = processor(e)
+    } else if (rxClass.test(key)) {
+      data.class = value
```
NozomuIkuta, line 42: CSS classes are supposed to be set at root level, according to [Vue documentation](https://vuejs.org/v2/guide/render-function.html#The-Data-Object-In-Depth), not as `data.attrs.class`.

NozomuIkuta:
By the way, according to [Vue documentation](https://vuejs.org/v2/guide/render-function.html#The-Data-Object-In-Depth), `style` attribute is also supposed to be set at root level of `data` object.

Should I fix it as well in this PR or submit another PR?

packages/content/templates/nuxt-content.js
```diff
@@ -53,7 +59,7 @@ function propsToData (node, doc) {
       obj[attribute] = value
     }
     return data
-  }, { attrs: {} })
+  }, { class: null, style: null, attrs: {} })
```
NozomuIkuta, line 62: It is good to set initial value `null`, for Vue not to render "empty" css class (e.g. `<p class=""></p>`).

packages/content/templates/nuxt-content.js
```diff
@@ -37,6 +39,10 @@ function propsToData (node, doc) {
       obj[field] = evalInContext(value, doc)
       data.on = data.on || {}
       data.on[event] = e => doc[value] = processor(e)
+    } else if (rxClass.test(key)) {
```
farnabaz, line 42: Creating an extra condition for `class` and `style` is redundant. Since there is `rootKeys` variable in [line 3](https://github.com/nuxt/content/blob/main/packages/content/templates/nuxt-content.js#L3) that detects root attributes like `class` and `style`.

The issue comes from attribute key. Remark create `class` attribute with `className` key and according to `rootKeys` array, `className` does not detects as root attribute. Adding `className` into `rootKeys` array in [line 3](https://github.com/nuxt/content/blob/main/packages/content/templates/nuxt-content.js#L3) will solve the issue.

```
 - const rootKeys = ['class-name', 'class', 'style']
 + const rootKeys = ['class-name', 'class', 'className', 'style']
``` 

As for the `style`, I believe it works as expected like `class` and there is no need to create custom logic for it.

packages/content/templates/nuxt-content.js
farnabaz: This is strange, Vue should not create empty `class` attribute when there is no value for it.  
Do you mind to create a simple reproduction for this issue?

NozomuIkuta: @farnabaz 

Here is a [demo](https://codesandbox.io/s/vue-class-attr-rendering-3hdc4), and here is [the related part in the official doc](https://v3.vuejs.org/guide/migration/attribute-coercion.html#attribute-coercion-behavior).

packages/content/templates/nuxt-content.js
NozomuIkuta: @farnabaz 

Thank you for pointing it out!
Yeah, actually I thought it was redundant, but couldn't come up with adding `className`. 😇 🙏 

I ran the test again and it passed (at least `component.test.js` since `highlighter.spec.js` has been failed recently)

packages/content/templates/nuxt-content.js
farnabaz: Thank you for the demo, but I mean a reproduction with nuxt-content.

I agree that having `class: ''` cause creating and empty attribute. 
Having null as default value for class and style is not necessary, and it should work without these defaults.

I suggest to remove these defaults and create another PR/issue with a reproduction sample that point the issue with nuxt-content

packages/content/templates/nuxt-content.js
farnabaz: Thanks 🙂

ManasMadrecha:
Any updates?

Atinux:
We are working with the Docus team to soon release the next version of @nuxt/content with few breaking changes and lot of fixes.

packet-sent:
Is the update going to happen after Nuxt 3 becomes a public beta?

packet-sent:
> 
> 
> We are working with the Docus team to soon release the next version of @nuxt/content with few breaking changes and lot of fixes.

Is the update going to happen after Nuxt 3 becomes a public beta?

packages/content/templates/nuxt-content.js
NozomuIkuta: @farnabaz 

I do appreciate if you take a look at my last comment and commit. 🙏 

> I pushed another commit to remove default values of class and style.

PR #909, opened 2021-06-14
fix(content): only use nuxt-link for absolute html links

### LINKED ISSUE #851: Opt out of automatic nuxt-link for linking to static content

All links that start with a '/' are replaced with nuxt-link. This makes it impossible to link to some file in the `static` directory.

There should be an option to opt out of autolinking, either as a new component, a tag, or a syntax modification to the link.

btkostner:
## Types of changes

- [X] Bug fix (a non-breaking change which fixes an issue)
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Description

This resolves #851 by only replacing links with `nuxt-link` if it's an absolute path without an extension. This was the most accurate way I could think of doing it, and should fix _most_ use cases.

No tests where added because I could not find anything related to the initial `a` to `nuxt-link` implementation. If you want tests, please direct me to where I should place those.

## Checklist:

### DIFF
--- a/packages/content/parsers/markdown/compilers/json.js
+++ b/packages/content/parsers/markdown/compilers/json.js
@@ -1,3 +1,5 @@
+const { extname } = require('path')
+
 /**
  * Parses nodes for JSON structure. Attempts to drop
  * unwanted properties.
@@ -13,10 +15,14 @@ function parseAsJSON (node, parent) {
     /**
      * Replace a tag with nuxt-link if relative
      */
-    if (node.tagName === 'a' && (node.properties.href || '').startsWith('/')) {
-      node.tagName = 'nuxt-link'
-      node.properties.to = node.properties.href
-      delete node.properties.href
+    if (node.tagName === 'a') {
+      const pathname = (node.properties.href || '')
+
+      if (pathname.startsWith('/') && extname(pathname) === '') {
+        node.tagName = 'nuxt-link'
+        node.properties.to = node.properties.href
+        delete node.properties.href
+      }
     }
 
     const filtered = {
--- a/packages/content/parsers/markdown/compilers/json.js
+++ b/packages/content/parsers/markdown/compilers/json.js
@@ -16,7 +16,7 @@ function parseAsJSON (node, parent) {
      * Replace a tag with nuxt-link if relative
      */
     if (node.tagName === 'a') {
-      const pathname = (node.properties.href || '')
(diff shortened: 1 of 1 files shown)


btkostner:
Hold up. Found a bug.

packages/content/parsers/markdown/compilers/json.js
```diff
@@ -13,10 +15,14 @@ function parseAsJSON (node, parent) {
...
      * Replace a tag with nuxt-link if relative
      */
-    if (node.tagName === 'a' && (node.properties.href || '').startsWith('/')) {
-      node.tagName = 'nuxt-link'
-      node.properties.to = node.properties.href
-      delete node.properties.href
+    if (node.tagName === 'a') {
+      const pathname = (node.properties.href || '')
+
+      if (pathname.startsWith('/') && extname(pathname) === '') {
```
farnabaz, line 21: Using `extname` for detect static files is not safe. For example returning value of `extname` for url `/setup#install-nuxt.js` is `.js`, which isn't correct and leads to invalid behavior.

Static detection should be done in markdown parser and ensure the file exists in `static` directory

btkostner: I'm confused about what you mean it should be done in the markdown parser. This is the markdown to json logic, and where it does the link modification...

As for using extname, it seemed like the fastest of two not great options. I can replace this with a `fs.existsSync` call instead but that would probably slow down the build time for large sites. It will be ran for every absolute link. Alternatively, I can strip out the hash from the path name, and still do the same extname logic.

farnabaz: Without checking file existence, this will be a breaking change. Since pages that have '.' in their names, will force full reload instead of client side navigation. 

As for the performance it can be improved by prefetching static files list.

btkostner: Yep!

Sadly it would require us to get the nuxt options to that function so we can use the configured static folder. Looking at the code, that shouldn't be too hard. Then we'd just need to watch the static folder for changes.

Considering the amount of work for that change, I'm going to hold off on writing it til I can get a Nuxt team member to comment on what they would like, as they are going to be the ones maintaining it.

Tahul:
Hello 🙂;

I think this issue has been fixed upstream in NuxtLink latest RFCs, I am so closing this as it should be fixed in v2.

Issue #932, opened 2021-07-27
The live editor expands by one row for each character typed


ElnuDev:
### Version
@nuxt/content: v1.14.0
nuxt: v2.15.7

In the [live editor](https://content.nuxtjs.org/displaying#live-editing), I'm getting this really weird issue where for every character I type, the text area expands by one row vertically, as you can see in [this video](https://cdn.discordapp.com/attachments/713346822143475752/869363375103164517/2021-07-26_16-38-33.mp4). Interestingly, it doesn't actually add extra lines to the markdown file, only expand the editor, but it's still extremely annoying.

PR #933, opened 2021-07-28
fix(content): prevent editor extending on typing

NozomuIkuta:
## Types of changes

## Description

Resolves: #932

In addition, I adjusted some CSS styles of textarea element.

## Checklist:

This is UI bug, so there is nothing to change in unit tests.
Instead, I tested it in Chrome (91.0.4472.164), Firefox (90.0.2), and Safari (14.1.1) on MacBookPro (Big Sur).

### DIFF
--- a/packages/content/templates/nuxt-content.dev.vue
+++ b/packages/content/templates/nuxt-content.dev.vue
@@ -98,7 +98,11 @@ export default {
 }
 
 .nuxt-content-editor {
+  box-sizing: border-box;
+  display: block;
   width: 100%;
   padding: 8px;
+  overflow: hidden;
+  resize: none;
 }
 </style>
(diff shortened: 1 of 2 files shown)


packages/content/templates/nuxt-content.dev.vue
```diff
@@ -98,7 +98,11 @@ export default {
 }
 
 .nuxt-content-editor {
+  box-sizing: border-box;
+  display: block;
   width: 100%;
   padding: 8px;
+  overflow: hidden;
+  resize: none;
```
NozomuIkuta, line 106: `width: 100% + horizontal padding` means the element overflows parents width.
So I added `box-sizing: border-box`.

Default style of `textarea` is `display: inline-block` but I explicitly set it as `block` (maybe this is just preference).

`overflow: hidden`  is to make situation simple and prevent redundant scroll bar from appearing.

`resize: none` is also to make situation simple. That is, with `overflow: hidden`, if developer shrink textarea vertically, overflown lines are hidden.

PR #971, opened 2021-10-04
Allow "as" prop to change the root div

chris-visser:
This allows developers to change the root `div` element of nuxt-context to something that might be semantically preferable like an `article` tag. This would avoid having to wrap nuxt-content with an element, saving some HTML.

### DIFF
--- a/docs/content/en/displaying.md
+++ b/docs/content/en/displaying.md
@@ -39,6 +39,8 @@ export default {
 - document:
   - Type: `Object`
   - `required`
+- as:
+  - Type: `String`
 
 Learn more about what you can write in your markdown file in the [writing content](/writing#markdown) section.
 
@@ -65,6 +67,14 @@ export default {
   }
 }
 </script>
+```
+
+## Root Element
+
+`<nuxt-content>` component will add a `div` element as the root of the content. You can change this by setting the `as` prop. Below example will use `article` as the root element.
+
+```vue
+<nuxt-content :document="doc" as="article">
 ```
 
 ## Style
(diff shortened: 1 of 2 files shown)


Atinux (commented):
Looks great! Do you mind adding a test case also?

chris-visser:
Will do!

Atinux (approved):

NozomuIkuta (commented):
@chris-visser 

Thank you for submitting the PR!

Yes, HTML semantics is important. 😄 

While logic itself LGTM as well, I wonder naming the prop `tag` instead of `as`.
Any thoughts?

My thought is based on...

- Vue Router 3.x provides `tag` prop to change the root element, so I believe people are familiar with the name (although it's removed from 4.x)
  - If I remember it correctly, other UI frameworks also use this term for same kind of prop.
- The prop is namely to change the "tag" of the root element
- The word "as" seems more like a keyword rather than prop name (at least for me), which is used when we convert/coerce something to another in (e.g. TypeScript types)
  - Off-topic: the word "as" is proposed as a Vue keyword to let props have different name eternally and internally (see vuejs/rfcs/pull/10), so it might become confusing in future

chris-visser:
@NozomuIkuta I agree. My idea was inspired by another library. Been a while so I forgot, but given that other Vue libs use or have used "tag", I'm fine to go along with that too.

chris-visser:
Done ty!

NozomuIkuta (approved):
@chris-visser 

Thank you for modifying the code!
all LGTM now. 🙋‍♀️

PR #999, opened 2022-01-10
fix: invalid default import from property-information

pbastowski:
The `property-information` package does not export a default object and thus `import info from "property-information"` returns `undefined`, resulting in the content module failing to render markdown files.

## Types of changes

## Description
I was trying to use the `nuxt-content` component to display a fetched markdown file, but it failed with an error. I followed the code and discovered that the import statement used in `nuxt-content.js` is incorrect. It tries to import a default `info` object from the property-information package, but there is no default object exported. Instead, all members are exported individually.

The fix involves using `import * as info from...` instead `import info from...`

## Checklist:

I have run tests and they all passed.

### DIFF
--- a/packages/content/templates/nuxt-content.js
+++ b/packages/content/templates/nuxt-content.js
@@ -1,4 +1,4 @@
-import info from 'property-information'
+import * as info from 'property-information'
 
 const rootKeys = ['class-name', 'class', 'className', 'style']
 


Tahul:
Hey; I am merging this as the solution seem valid, this might be a part of the last `v1` version we release.

PR #1052, opened 2022-05-10
fix: use builtin storage to watch contents

### LINKED ISSUE #1036: Use built-in storage
Starting with rc.3 and nitro 0.4.x, you can access storage instance with the same configuration from `nitro.storage` (using `nitro:init` hook to access) and to replace [custom instance in module](https://github.com/nuxt/content/blob/19c8142e8642a31df63229835d7452476c461ec8/src/module.ts#L379) and unstorage dependency. /cc @farnabaz

farnabaz:
### 🔗 Linked issue
resolve #1036

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -10,7 +10,6 @@ import {
   useLogger
 } from '@nuxt/kit'
 import defu from 'defu'
-import { createStorage } from 'unstorage'
 import { join } from 'pathe'
 import type { Lang as ShikiLang, Theme as ShikiTheme } from 'shiki-es'
 import { listen } from 'listhen'
@@ -19,7 +18,6 @@ import { debounce } from 'perfect-debounce'
 import { name, version } from '../package.json'
 import {
   createWebSocket,
-  getMountDriver,
   MOUNT_PREFIX,
   processMarkdownOptions,
   PROSE_TAGS,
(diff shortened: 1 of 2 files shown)


src/module.ts
```diff
@@ -374,39 +372,33 @@ export default defineNuxtModule<ModuleOptions>({
...
 
-    // Dispose storage on nuxt close
-    nuxt.hook('close', async () => {
-      await Promise.all([
-        storage.dispose(),
-        ws.close()
-      ])
-    })
+      // Dispose storage on nuxt close
+      nuxt.hook('close', async () => {
```
pi0, line 379: i think you can also use `nitro.hooks.hook('close')` since we are in nitro context and was depends on nitro storage

src/module.ts
```diff
@@ -374,39 +372,33 @@ export default defineNuxtModule<ModuleOptions>({
...
+        await ws.close()
+      })
 
-    // Listen dev server
-    const { server, url } = await listen(() => 'Nuxt Content', { port: 4000, showURL: false })
-    server.on('upgrade', ws.serve)
+      // Listen dev server
+      listen(() => 'Nuxt Content', { port: 4000, showURL: false })
```
pi0, line 384: We should avoid dangling promises. await should be fine

src/module.ts
```diff
@@ -374,39 +372,33 @@ export default defineNuxtModule<ModuleOptions>({
...
+          const broadcast = debounce((event: WatchEvent, key: string) => {
+            key = key.substring(MOUNT_PREFIX.length)
+            logger.info(`${key} ${event}d`)
+            ws.broadcast({ event, key })
+          }, 50)
 
-    // Watch contents
-    storage.watch(broadcast)
+          // Watch contents
+          nitro.storage.watch(broadcast)
```
pi0, line 399: `watch` also returns a promise we need to await

PR #1073, opened 2022-05-12
feat!: Refactor for a great DX

Atinux:
## ❓ Type of change

### API Endpoints

`/api/_content/query` and `/api/_content/navigation` now works and you can use it with a simpler query to test in development:

- http://localhost:3000/api/_content/query?path=/&only=title
- http://localhost:3000/api/_content/query?sortBy=size:desc
- http://localhost:3000/api/_content/query?without=body

### AST

It makes more sense from my POV to use `path` instead of `slug` (usually a slug does not have any `/` in it).

The changes are:
- `slug` becomes `path`
- `path` becomes `file`

This will bring breaking changes for Docus but are worth it for a clearer interface returned by Content.

### `<Content path="/hello" />`

Fetch the document and display it, it does not accept a `document` argument anymore but instead a `path`.

If no `path` is defined, it will fallback to: `useRoute().path`.

```html
<Content /> <!-- Similar to <Content :path="$route.path" /> -->
```

If the path is currently the same as `useRoute().path`, the component will also set the page title and meta description based on its front-matter. You can also use the `head` property in the document front-matter.

Bonus, it can be used as well inside the Markdown to include partials:

```md
:content{path="/_partials/beta-alert"}
```

This means, **a minimal Nuxt Content project can be as simple as**:

`app.vue`
```vue
<template>
  <Content :key="$route.path" />
</template>
```

Or if in `pages/[...slug].vue`
```vue
<template>
  <Content />
</template>
```

TODO:
- [ ] Slots

### `<Document />`

This was `<Content />` before but now uses `v-model` instead of a `document` prop:

```vue
<script setup>
const { data: document } = await useAsyncData('hello', () => queryContent('/hello').findOne())
</script>

<template>
  <Document v-model="document" />
</template>
```

## 📝 Checklist

- [ ] I have updated the documentation accordingly
- [ ] A PR to Docus has been made to change `slug` to `path`


docs/content/1.getting-started/2.usage.md
```diff
@@ -1,79 +1,67 @@
...
+### CSV
+
+### JSON(5)
+
+
+
 ## Display the content
 
-Edit the `app.vue` file to query your content with the auto-imported `queryContent` composable. The `/` path corresponds to `content/index.md` (root of the directory).
+Anyware in your Nuxt application, you can use the [`<Content>`](/api/components#content) component to render the content of a file matching the current route.
```
clemcode, line 54:
```suggestion
Anywhere in your Nuxt application, you can use the [`<Content>`](/api/components#content) component to render the content of a file matching the current route.
```

src/runtime/types.d.ts
```diff
@@ -38,6 +38,13 @@ export interface ParsedContentMeta {
 }
 
 export interface ParsedContent extends ParsedContentMeta{
+  /**
+   * Exceprt
```
clemcode, line 42:
```suggestion
   * Excerpt
```

farnabaz:
Test are updated, but Netlify fails because it tries to deploy content v1 website instead of v2?

Tahul:
I have commit ready to update Docus to this PR :)

Waiting for publish to push it!

With latest commits fixing tests, it looks good to me!

Really nice improvement! ❤️ 

```vue
<template>
   <Content />
</template>
```

☝️ this looks lovely

Tahul:
(merging; docs deployment will break until I push the update of Docus)

PR #1088, opened 2022-05-16
feat(components)!: update and improve components

Tahul:
Hey 👋 

This PR improve components provided by Nuxt Content:

- `<Content />` is now `<ContentDoc />` with new props:
	- `query`: allows to pass a QueryBuilderParams
	- Will use MarkdownRenderer if not supplied with a default slot:
	   ```vue
       <template><ContentDoc /></template>
	- can now be used as renderless component:
	  ```vue
	  <template>
	  <Content path="/my/file/path" v-slot="{ data }">
	    {{ data.title }}
	  </Content>
	  </template>
	  ```
	  - now has `empty` slot
	  ```vue
	  <template>
	  <Content path="/not-found" v-slot="{ data }">
	    <template #empty>
	  	  No content found!
	    </template>
	  </Content>
	  </template>
	  ```

- `<Document />` becomes `<ContentRenderer />`
	- [`<document />`](https://developer.mozilla.org/fr/docs/Web/API/Document) element is a native HTML element
	- `document` naming is ambiguous as `document` is a global in browser environments.
	- also supports `renderer` props, and supports renderless usage as well.
	- also supports `#empty` slot
 
- adds new `<ContentList />`
   - Works same as `<ContentDoc>` except it does not uses MarkdownRenderer
   - You can access data this way: `<ContentList path="/your/content/dir/path" v-slot="{ list }" />`
 
- adds new `<ContentNavigation />`
	-  accepts `query` similar as `<ContentDoc />`
	```vue
	<template>
	  <ContentNavigation v-slot="{ navigation }">
	    {{ navigation }}
	  </ContentNavigation>
	</template>
	```

I moved components implementation to `.ts` files as I think it's better to track missing dependencies and exposed typings as auto-imports ain't working yet for `node_modules`!

I also think it's better to ensure consistency accross renderless components implementations, and might also be easier to type slot-scoped data!

 
I updated documentation accordingly, you can check it out in API / Components 🙂  

As component name changed and Docus can't be upgraded yet, until this gets merged you'll have to update this to run documentation:

 ```
 // docs/node_modules/@docus/docs-theme/pages/[...slug].vue
 // Replace <Document /> by <ContentRenderer />
 ```


docs/content/3.api/5.components/1.content.md
```diff
@@ -0,0 +1,104 @@
...
+  </main>
+</template>
+```
+
+### Filling components with JSON files
+
+```vue [pages/[...slug.vue]]
+<template>
+  <main>
+    <Content path="/article/authors" v-slot="{ data }">
```
Atinux, line 97: Shoudn't it be with `type="find"` ?

farnabaz:
Nice work @Tahul  👍 

I'm thinking about the `Content` components, I think instead of having type `props` it will be good if we  create separate component `ContentList` to list contents.
Having separate component is makes more sense to me, because I can detect What I’m gonna expect from it by just reading its name.
Also `type` prop might be confusing for new users.

pi0:
This is a really nice idea!

On small note: I would strongly avoid using single part components since is against Vue component naming convention. `<ContentQuery>`, `<ContentDocument>` (shortcut with slug find) and `<ContentList>` might be good ideas.

Regarding consistency with Framework, we are planning to introduce data-fetching shortcuts such as `<NuxtAsyncData>` and `<NuxtFetch>`. I believe they would be rootless components as data-provider without a wrapper tag (like in current implementation). 

It would be much better if you could somehow decouple renderer part from data provider wrapper using a nested component like this:


```html
<!-- Renders nothing -->
<ContentDocument slug="/foo/bar" v-slot={ data }>
   Blog article:
      <ContentRender :data="data" tag="p" />
</ContentDocument>
```

Benefits are:
- It makes data provider pattern same as Framework ones without wrapper element
- It makes usage as much as customizable as possible
- It allows using abstract data to render custom content (non markdown for rendering lists, etc)
- It reduces chances of breaking changes 
- It allows combining Vue code for adding some other elements inside the wrapper

Tahul:
> This is a really nice idea!
> 
> On small note: I would strongly avoid using single part components since is against Vue component naming convention. `<ContentQuery>`, `<ContentDocument>` (shortcut with slug find) and `<ContentList>` might be good ideas.
> 
> Regarding consistency with Framework, we are planning to introduce data-fetching shortcuts such as `<NuxtAsyncData>` and `<NuxtFetch>`. I believe they would be rootless components as data-provider without a wrapper tag (like in current implementation).
> 
> It would be much better if you could somehow decouple renderer part from data provider wrapper using a nested component like this:
> 
> ```
> 
> <ContentDocument slug="/foo/bar" v-slot={ data }>
>    Blog article:
>       <ContentRender :data="data" tag="p" />
> </ContentDocument>
> ```
> 
> Benefits are:
> 
> * It makes data provider pattern same as Framework ones without wrapper element
> * It makes usage as much as customizable as possible
> * It allows using abstract data to render custom content (non markdown for rendering lists, etc)
> * It reduces chances of breaking changes
> * It allows combining Vue code for adding some other elements inside the wrapper

Hello Pooya;

That pattern is exactly what is implemented currently in that PR.

`<Content>` component is in charge of querying feature, `<ContentRenderer>` is in charge of rendering the content.

You can also use `<Content>` as a standalone component by disabling the renderer usage:

```
<Content :renderer="false" v-slot="{ data }">
   {{ data.title }}
</Content>
```

Tahul:
> Nice work @Tahul 👍
> 
> I'm thinking about the `Content` components, I think instead of having type `props` it will be good if we create separate component `ContentList` to list contents. Having separate component is makes more sense to me, because I can detect What I’m gonna expect from it by just reading its name. Also `type` prop might be confusing for new users.

Thank you a lot, that is a really nice suggestion which will clarify the usage from <template> a lot, will be pushing that change right now. 🙂

pi0:
We can create `<ContentDocument>` that internally combines `<ContentQuery>` same API (potentially even reusing) of framework core and `<ContentRenderer>` if you like to make shortcut wrapper and believe it is the most common usage still preserve modularity :)

---

Off topic: With Nuxt3, we are avoiding multiple names for same component (such as deprecated NLink) You might consider this for introducing aliases.

Tahul:
> We can create `<ContentDocument>` that internally combines `<ContentQuery>` same API (potentially even reusing) of upcoming framework data provider components and `<ContentRenderer>` if you like to make shortcut wrapper and believe it is the most common usage for markdown still preserve modularity :)
> 
> Off topic: With Nuxt3, we are avoiding multiple names for same component (such as deprecated NLink) You might consider this for introducing aliases.

That is a nice suggestion, I was looking at having `<ContentQuery />` as a base for `<ContentDocument />` and `<ContentList />`.

Will go that way 🙂

Atinux:
Let's try this :+1:

farnabaz (approved):
👍

Atinux (approved):

PR #1090, opened 2022-05-16
feat(sort): replace `sortBy()` with advanced `sort()`

### LINKED ISSUE #1074: V2: Add new query `SortByNatural`
Currently (even in V1), we can only fetch and sort using normal `SortBy`, but it sorts files like this... 1-name, 10-name, 2-name, 3-name, etc. 

1. To be able to sort the files naturally like this... 1-name, 2-name, 3-name, 10-name
2. So, in addition to the `SortBy` option, kindly add one more option of `SortByNatural` using `Intl.collator`
    https://github.com/nuxt/content/blob/1968257881db58634bdf4448bd85d7b059f88e4c/src/runtime/query/query.ts#L35

This way we can directly naturally sort while fetching itself and then properly use `limit`, etc. options. 

Fetch thousands of posts without any sort, and then use `Intl.collator` to sort them, and then filter out only first n posts. But it's not performant if there are so many posts. 

### Additional context
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator

farnabaz:
### 🔗 Linked issue
resolves #1074

### ❓ Type of change

### 📚 Description

BREAKING CHANGE: drop `sortBy` in favor of new `sort`

### 📝 Checklist

### DIFF
--- a/docs/content/3.api/2.query-content.md
+++ b/docs/content/3.api/2.query-content.md
@@ -63,38 +63,41 @@ const products = await queryContent('products').where({ 'categories': { $contain
 const products = await queryContent('products').where({ 'categories': { $contains: ['top', 'woman'] } }).find()
 ```
 
-## `sortBy(field, direction)`
+## `sort(options)`
 
-- `field`{lang="ts"}
-  - Type: `string`{lang="ts"}
+- `options`{lang="ts"}
+  - Type: `object`{lang="ts"}
   - `required`{lang="ts"}
-- `direction`{lang="ts"}
-  - Type: `string`{lang="ts"}
-  - Value: `asc | desc`{lang="ts"}
-  - Default: `asc`{lang="ts"}
 
-Sort results by a field.
+Sort results by a field or fields.
 
 ```ts
 // Sort by title ascending
 const articles = await queryContent('articles')
-  .sortBy('title')
+  .sort({ title: 1 })
   .find()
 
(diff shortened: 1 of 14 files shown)


Atinux:
Do you mind adding more tests for the sensitivity?

docs/content/3.api/2.query-content.md
```diff
@@ -63,38 +63,41 @@ const products = await queryContent('products').where({ 'categories': { $contain
...
+  .sort({ 'category.title': 1 })
   .find()
 
 ```
 
-> `sortBy()`{lang="ts"} method does **case-sensitive** sort.
+> `sort()`{lang="ts"} method does **case-sensitive** sort by default. There is some magical options you can pass to sort options to change sort behavior, like sorting **case-insensitive**.
+> - `$sensitivity`: Change case sensitivity. Like using `$sensitivity: 'base'` for case-insensitive sort
+> - `$numeric`: Whether numeric collation should be used, such that `"1" < "2" < "10"`.
+> - `$caseFirst`: Whether upper case or lower case should sort first.
```
Atinux, line 100:
```suggestion
> - `$caseFirst`: Whether upper case or lower case should sort first.

These options are given to [Intl.Collator()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#parameters)
```

PR #1094, opened 2022-05-17
feat: preview mode

### LINKED ISSUE #1067: Content Live Preview
The idea is to support live preview in deployed project. Meaning that we should be able to  inject  preview contents into the content storage on runtime in production.

Preview contents can:
- Add new files to content storage (like new pages)
- Edit an existed contents and overwrite them
- Delete an existed content and make it unavailable

As we talked with @pi0, here is steps we should take:

- [ ] Have a standard way to detect, persist and get preview key on server side

```ts
isPreview(event)
getPreview(event)
```

- [ ] Rename server side `queryContent(params)` into `serverQueryContent(event, params)` in order to extract detect preview mode and extract preview key

- [ ] Detect preview mode on content listing and fetching, attempt to load content from preview storage, and fallback to cache storage.

!Preview storage is a specific prefix(`preview:`) under `sourceStorage`. To found updated version of content in production we can do like this:

```ts
const { key } = getPreview(event)
const previewId = 'preview:' + key + ':' + original_content_id
const previewContent = sourceStorage.getItem(previewId)
```

farnabaz:
### 🔗 Linked issue
resolve #1067

### ❓ Type of change

### 📝 Checklist


docs/nuxt.config.ts
```diff
@@ -13,6 +13,7 @@ export default defineNuxtConfig({
   generate: {
     routes: []
   },
+  modules: ['@nuxthq/admin'],
```
Atinux, line 16: Is it required since used in Docus base?

```suggestion
```

src/runtime/server/preview.ts
```diff
@@ -0,0 +1,13 @@
+import type { CompatibilityEvent } from 'h3'
+import { useQuery, useCookie } from 'h3'
+
+export const isPreview = (event: CompatibilityEvent) => {
+  const previewToken = useQuery(event).previewToken || useCookie(event, 'previewToken')
```
Atinux, line 5: Can we use `getPreview(event)` here to simplify?

src/runtime/server/preview.ts
```diff
@@ -0,0 +1,13 @@
...
+  return !!previewToken
```
Atinux, line 6:
```suggestion
  return !!getPreview(event)
```

Atinux:
Merging and iterating later on this 😊

PR #1097, opened 2022-05-18
feat: new merging strategy

Atinux:
### ❓ Type of change

### 📚 Description

The idea behind this ongoing PR is to start prefixing the special keys with a `$` to allow a merging strategy with the file content (Markdown metadata, YML & JSON).

This allows to easily use the `where()`, `only()` and `without()` operators without the need to always prefix with `body.` and also make things similar to Content v1.

```json
{
  "$atime": "2022-05-18T07:12:09.280Z",
  "$mtime": "2022-05-18T07:01:56.895Z",
  "$size": 33,
  "$path": "/titi",
  "$draft": false,
  "$partial": false,
  "$empty": false,
  "$type": "markdown",
  "$id": "content:toto.md",
  "$source": "content",
  "$file": "toto.md",
  "$extension": "md",
  "title": "Toto 2",
  "description": "",
  "body": { ... },
}
```

This will change the query to use `where({ $path: '/titi' })` but that makes sense from my POV since those are generated keys (similar to MongoDB with the `_id`).

We can also introduce a `withoutMeta()` method to remove all the meta keys, or directly supporting `without('$')` ?

### 📝 Checklist


Atinux:
I would personally remove the $mtime, $atime since when deploying they are not respected and the $size is not that helpful here (mostly useful for hash generation for caching)

The $extension as well since we have the $type

farnabaz:
What do you think about using `_` instead of `$` for prefixing? (like mongo)
Query builder use `$` prefix to distinct between fields and operators, and using `$prefix` for fields my create conflict,  for query readability and also for searching behavior.

We already have a `$type` operator in query builder which checks type of specific field.

```
queryContent().where({
	someField: { $type: 'number' }
})
```

Atinux:
I am in between since the `$type` or other are applied only after the key definition, so it would work still:

```ts
queryContent().where({
	$size: { $gte: 10 }
})
```

vs 

```ts
queryContent().where({
	_size: { $gte: 10 }
})
```

farnabaz:
Personally I like to separate these concepts visually, keep `$` as preserved key for query operations just like mongo and use `_` for internal fields.

Query builder and operation are not full featured yet, having strict rule over it will help us to improve it without having issue or conflict in future.

Tahul:
Please could you add `withoutMeta()` or any other sanitizing function to this PR?

We definitely need this to ensure we won't be maintaining these ourselves in content projects.

Atinux:
I will yes @Tahul

farnabaz (approved):

PR #1099, opened 2022-05-18
feat(cache): respect module version

### LINKED ISSUE #1098: Add Nuxt Content Version in cache storage (dev)
This way we can flush the cache content in dev if we detect a different version (in case we add new fields for example)

farnabaz:
### 🔗 Linked issue
resolves #1098
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -385,7 +385,10 @@ export default defineNuxtModule<ModuleOptions>({
       wsUrl: ''
     })
     // Context will use in server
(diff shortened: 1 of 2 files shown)


Atinux (approved):

pi0:
This change also wrongly invalidates cache when not necessary! You could preserve an internal version that represents cache format independent from module versioning (example: https://github.com/unjs/jiti/blob/main/src/jiti.ts#L32)

PR #1105, opened 2022-05-19
feat: improve markdown parser and add playground

farnabaz:
### ❓ Type of change

### 📝 Checklist


docs/pages/playground.vue
```diff
@@ -0,0 +1,110 @@
+<script setup>
+import { parse } from '../../src/runtime/markdown-parser'
```
Atinux, line 2: Would that work when deploying?

farnabaz: Yes, it will bundled in client chunks

Tahul:
Playground is almost ready, I just don't really know how to ship/bundle monaco-editor to be available in production.

Would love some help on that.

docs/nuxt.config.ts
```diff
@@ -21,6 +21,17 @@ export default defineNuxtConfig({
   github: {
     repo: 'nuxt/content'
   },
+  vite: {
+    define: {
+      'process.env': {}
```
pi0, line 26: https://github.com/nuxt/content/pull/1105/files#diff-64109d102581711f2d242e8db110546964e19787939bc6a85d3f7a83104db518R1

image

docs/components/content/Playground.vue
```diff
@@ -1,52 +1,83 @@
 <script setup>
-const PARSE_SERVER = 'https://mdc.nuxt.dev/api/parse'
+import { parse } from '../../../src/runtime/markdown-parser'
```
pi0, line 2: Including server-related logic in the app, the runtime is a bad idea and anti-pattern, even for docs and playground, and by lazy loading, we are making total bundle size at least 3x! (same parser to server/SSR/Client). 

We could use a local API with the keep-alive connection that reduces delay or even service-worker target of nitro, making it zero latency.

Atinux: I personally don't see an issue since it's loaded on-demand:

https://user-images.githubusercontent.com/904724/169517170-113d5fb6-34f8-4db4-afa2-9c618f27dd90.mp4


Also the API for it is bad in term of UX (we tried with https://mdc.nuxt.dev) and consume unnecessary server resources and cost.

pi0: You just downloaded 8MB in the above recording + Hidden overhead on the browser to load all of this :) As I mentioned, we could benefit from SW target if you concern about hosting costs...

Tahul: Hey :)

Using Service Worker was our first objective for this playground, but no chance during implementation.

Could you maybe point us on examples or instructions on how to do it?


Would love to get help on these points as mentioned in the commit above.

Atinux: > You just downloaded 8MB in the above recording + Hidden overhead on the browser to load all of this :) As I mentioned, we could benefit from SW target if you concern about hosting costs...

Happy to improve it on another PR but the benefits of the playground in term of documentation is a must have and worth the extra MB (which are cached for future navigation)

pi0: I would be more than happy to help make the service worker idea work. It is probably even already possible with nitro preset will try on a small demo! But it needs implementation to still be in the server and changes in this PR is against it!

src/runtime/markdown-parser/frontmatter.ts
```diff
@@ -1,6 +1,8 @@
-import matter from 'gray-matter'
```
pi0, line 1: Regardless of playground discussion, I believe parser is normally used in server and can be moved to `runtime/server/markdown-parser` for code clarity.

docs/components/content/Playground.vue
Tahul: Most likely this can be started from another branch (`main` still has server-only `parse`), and merged to this one if this is reliable solution.

I know @farnabaz tried to use SW from Nitro without chance.

pi0: Sure! As long as we are in the same context that this usage in playground needs to be reverted, I'm fine to not block the changes. Will provide you an example repo. BTW you can always DM me when having issues ;)

docs/nuxt.config.ts
```diff
@@ -23,7 +23,10 @@ export default defineNuxtConfig({
   },
   vite: {
     define: {
-      'process.env': {}
+      'process.env.FORCE_COLOR': {},
```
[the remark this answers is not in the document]
Tahul: I don't really know why it is bundled, we are willing to move to CDN usage for Monaco anyways!

src/runtime/markdown-parser/frontmatter.ts
Atinux: Actually I think we should have the ability to make it run on client-side as well (ideally through a web worker)

docs/components/content/Playground.vue
Atinux: Still I don't believe SW is the solution here, I would use a Web Worker to load this code and execute it to not block the main thread (even though it is quite fast actually)

src/runtime/markdown-parser/frontmatter.ts
pi0: What would be the use-case and benefits?

docs/components/content/Playground.vue
pi0: Would be happy to discuss about benefits of web-worker approach :)

Tahul:
Hey :)

I merged this PR so we can profit of from the actual playground for the release, but would be really happy to continue the discussions in this follow-up issue: #1110 

Good job for all the work on this PR, I'm so happy that we now have our own playground! 😄

PR #1114, opened 2022-05-23
feat: internal cache version

farnabaz:
### ❓ Type of change

### 📚 Description

> You could preserve an internal version that represents cache format independent from module versioning
https://github.com/nuxt/content/pull/1099#issuecomment-1132829442

Introduce `cacheVersion` config to invalidate content caches instead of using module version.

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -18,6 +18,7 @@ import { listen } from 'listhen'
 import type { WatchEvent } from 'unstorage'
 import { name, version } from '../package.json'
 import {
+  CACHE_VERSION,
   createWebSocket,
   MOUNT_PREFIX,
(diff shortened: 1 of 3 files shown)


src/module.ts
```diff
@@ -18,6 +18,7 @@ import { listen } from 'listhen'
 import type { WatchEvent } from 'unstorage'
 import { name, version } from '../package.json'
```
Atinux, line 19:
```suggestion
import { name } from '../package.json'
```

farnabaz: Version is used in package `meta` https://github.com/nuxt/content/blob/feat/internal-cache-version/src/module.ts#L173

PR #1116, opened 2022-05-23
feat(markdown): web-friendly excerpt delimiter (case & space insensitive)

Byloth:
`RegExp` are also allowed.

### ❓ Type of change

### 📚 Description

This PR allows the developer to define which `delimiter` can be used to identify a document excerpt via a simple configuration.

It also allows the developer to use a `RegExp` object instead of a simple `string`.  
Both syntaxes are supported.

### 📝 Checklist

At the moment, I can't check these boxes; I didn't create any Issue of Discussion about it.  
Let me know if it's really necessary or if in this simple case, a PR may be enough.

I was also unable to find some documentation, about `excerpt` and its use, in this repository.  
Is it located elsewhere?


Tahul:
Hey 👋 ,

Thank you for the contribution, this seem like an interesting idea!

Could you please explain your personal use case for this? I'm interested!

I will let @farnabaz do a second check but these changes looks good to me!

Tahul (approved):

farnabaz (changes requested):
Thank you @Byloth for the PR.

For now it's not possible to pass Regex from `nuxt.config` into runtime config. Let's drop RegExp support for delimiter and stick with simple string for now.

Atinux:
Agree with @farnabaz to keep a simple string, Regex is not necessary for such feature.

Atinux:
It would be nice also to add a section in Markdown regarding the `excerpt` feature: https://content-v2.nuxtjs.org/guide/writing/markdown

Byloth:
Hi, everybody.  
Thanks for all your reviews.

> Could you please explain your personal use case for this?

It's just a feature to help the user avoid making mistakes.  
In my experience, when you need to explain a thing (as in `@nuxt/content` documentation: *«Be careful to enter  exactly; i.e., all lowercase and with no whitespace.»*) it means that something is wrong... Especially in the Web context, where everything is case-insensitive and spaces are hardly relevant at all.  

Of course, it also has a real use case...
Until now, I have pre-parsed everything my users wrote, by looking for all matches of `/<!--\s*more\s*-->/i` RegExp and brutally replacing them with the standard `<!--more-->` delimiter... As you can deduce, it isn't a very elegant solution.

---

> For now it's not possible to pass Regex from `nuxt.config` into runtime config.

Could it be a solution to define an extended type support for `excerptDelimiter` like:

```ts
interface RegExpOptions
{
  pattern: string | RegExp
  flags?: string
}
interface MarkdownOptions
{
  // [...]
  excerptDelimiter: string | RegExpOptions
  // [...]
}
```

... allowing the developer to specify in `nuxt.config` something like:

```jsonc
// [...]
    excerptDelimiter: "<!-- something -->",
    
    // ... or...
    
    excerptDelimiter {
      pattern: "<!--\\s*more\\s*-->",
      flags: "i"
    }
// [...]
```

... and then use the `pattern` key (if present) to instantiate a new `RegExp` object like:

```ts
// [...]
    excerptDelimiter = options.excerptDelimiter
    if (excerptDelimiter.pattern)
      excerptRegex = new RegExp(excerptDelimiter.pattern, excerptDelimiter.flags)

// [...]
```

Could it be acceptable?  
Can I implement this and update this PR?

farnabaz:
> Could it be a solution to define an extended type support for excerptDelimiter like

Introducing new syntax to support regex seems wrong to me. 
This will leads us to trouble, specially when Nuxt support Regex in runtime config and we need to deprecate this syntax.

A solution to support Regex could be:
-  Convert RegExp to string with `.toString` 
- Convert string representation of regex back to RegExp in markdown transformer. (We can check regex validity using `/\/(.*)\/([dgimsuy]*)$/`)

WDYT @Atinux ?

Atinux:
I am not fan of the idea of giving a configuration for the excerpt directly.

BUT, I believe having a more case-insensitive excerpt is much better in term of UX/DX.

Can we simply update our delimiter to be a regex directly? https://github.com/nuxt/content/blob/docs/marketing/src/runtime/markdown-parser/index.ts#L84

farnabaz:
I agree, internally update excerpt extractor to use regex and remove custom delimiter from module options.

Atinux:
This will also avoid complexity for the online editor :blush:

Byloth:
> Introducing new syntax to support regex seems wrong to me. This will leads us to trouble, specially when Nuxt support Regex in runtime config and we need to deprecate this syntax.

Mmmh... Ok, ok!
Got it! 🤔

---

> Can we simply update our delimiter to be a regex directly? 

Yeah, of course!

Developers haven't had the chance to customize it until now anyway...  
At least, it will support a more web-friendly syntax helping to avoid mistakes! 🙃

LGTM!

---

I going to update this PR...

Atinux (approved):

Byloth:
Oh...  
I wished to contribute also adding some documentation, about the `excerpt` and its use... 😅

No problem...
I'm going to open a new PR! 😉

Atinux:
Yes feel free to open a new PR :blush:

PR #1145, opened 2022-05-25
docs: added information on sitemap generation

heychazza:
### 🔗 Linked issue
https://github.com/Atinux/content-wind/issues/5

### ❓ Type of change

### 📚 Description

Sometimes users may want to generate a sitemap file, to make it easier for Google to index the content within your website. I personally had this issue, and after a discussion with @Atinux within https://github.com/Atinux/content-wind/issues/5, he felt this could be good addition to the docs.

### 📝 Checklist

### DIFF
--- a/docs/content/3.guide/2.displaying/5.sitemaps.md
+++ b/docs/content/3.guide/2.displaying/5.sitemaps.md
@@ -1,0 +1,36 @@
+---
+title: Sitemaps
+---
+
+A sitemap file is useful for helping Google to better index your website, ensuring that the content you write can be visible on search results.
+
+This can be created utilising the `sitemap` library, so you'll need to install that which can be done like so:
+  ```bash
+  yarn add sitemap
+  ```
+
+## Server Route
+We will be utilising the server routes available within Nuxt 3, and to do so you'll need to create the `server/` folder within your websites root directly.
+
+Once this is done, create a `routes/` folder inside this, and add a `sitemap.xml.ts` file, this corresponds to `https://example.com/sitemap.xml`.
+
+You'll need to add the following:
+
+```yaml [server/routes/sitemap.xml.ts]
+import {serverQueryContent} from '#content/server';
+import {SitemapStream, streamToPromise} from 'sitemap';
+
+export default defineEventHandler(async (event) => {
+    const articles = await serverQueryContent(event).find();
+
+    const sitemap = new SitemapStream({ hostname: 'https://example.com' });
+    articles.forEach((article) => sitemap.write({ url: article._path, changefreq: 'monthly' }));
+    sitemap.end();
+
+    let data = await streamToPromise(sitemap);
+    return data;
+});
+
+```
+
+Now, once users go to `https://example.com/sitemap.xml`, you'll find the generated XML file with all your blog posts.


heychazza:
Just realised those using static sites won't be able to utilise this.. so solely only for those with SSR.

Tahul:
Hello @heychazza :)

That looks like a nice addition to the docs!

Do you think it could be worth to implement this feature natively in @nuxt/content?

We already have navigation generation, so I guess it would not be so hard to create a sitemap out of this.

Maybe we could collaborate on this PR towards a native feature with its own documentation page?

cc @Atinux @farnabaz

Tahul:
Hello @heychazza :)

That looks like a nice addition to the docs!

Do you think it could be worth to implement this feature natively in @nuxt/content?

We already have navigation generation, so I guess it would not be so hard to create a sitemap out of this.

Maybe we could collaborate on this PR towards a native feature with its own documentation page?

cc @Atinux @farnabaz

heychazza:
> Hello @heychazza :)
> 
> 
> 
> That looks like a nice addition to the docs!
> 
> 
> 
> Do you think it could be worth to implement this feature natively in @nuxt/content?
> 
> 
> 
> We already have navigation generation, so I guess it would not be so hard to create a sitemap out of this.
> 
> 
> 
> Maybe we could collaborate on this PR towards a native feature with its own documentation page?
> 
> 
> 
> cc @Atinux @farnabaz

Hey buddy, I think this would be a superb addition to give people out-the-box sitemap generation for their blogs.

farnabaz:
Generating sitemaps mostly is project specific thing and generalizing it in content module may create some edge case issues. 
- How we should treat partials? ignore then or assume they are standalone page? (if we think of them as a page we should have a rule to exclude `_dir`, `_theme` files)
- What should we do with non-markdown files?
- Users should be able to include/exclude contents in sitemap (also include their custom pages under `pages/` directory)

Having documentation about how to create sitemap seems good to me, but we should have some notes about the partials, non-markdown and content filtering.

Atinux:
I don't think adding it into Content features is worth it like @farnabaz said, it is specific to each website and an example how to achieve it best from my POV.

I would maybe create a new category for these features, what about "Recipes"?

We could later on showcase also how to leverage the scheme-org module by @harlan-zw for instance.

heychazza:
> I don't think adding it into Content features is worth it like @farnabaz said, it is specific to each website and an example how to achieve it best from my POV.
> 
> I would maybe create a new category for these features, what about "Recipes"?
> 
> We could later on showcase also how to leverage the scheme-org module by @harlan-zw for instance.

How about the idea of having a module for it? I know @nuxt/sitemap exists, but seems like content v2 isn’t yet supported

Atinux:
Let's start first with an example in the documentation and get feedback from it first ☺️

Atinux:
For sitemap generation when using `nuxt generate`, we can advice to push the `/sitemap.xml` route into the `nitro.prerender` option in `nuxt.config`

harlan-zw:
Recipes sound good. 

I'm sure we'll also see a bunch of specialised boilerplates and tutorials popping up which will also help users get set up for common use cases.

Also, last I heard on the sitemap module was that nitro would handle this as a plugin, so recommending bespoke sitemap generation may not be helpful once it's released.

Atinux:
Well we will be able to update the example once Nitro offers a better way to do it :)

benoitdemaegdt:
Hey there 👋 

I'm a bit late on this PR since it is already merged 😅 
However I was wondering why did you choose to use a server route for generating a sitemap and not a module ?

[This](https://github.com/benoitdemaegdt/nuxt3-sitemap#setup-for-a-dynamic-site-powered-by-nuxtcontent-with-prerendering) is an attempt to generate a sitemap for dynamic sites powered by @nuxt/content and prerendered. So far it seems to be working even if code may be improved for better clarity.

I would be happy to talk about it and share progress if you are interested 😄

Atinux:
Nice work @benoitdemaegdt 

It would be nice to update the structure to use the module starter for best packaging: https://github.com/nuxt/starter

PR #1200, opened 2022-06-02
feat(types): provide augmentations for `only` and `without`

harlan-zw:
### ❓ Type of change

### 📚 Description

Allows type hinting on the keys for `only` and `without` as well as augmenting the result which will either include or exclude the provided keys from the IntelliSense respectively.

image

image

### 📝 Checklist

### DIFF
--- a/src/runtime/query/query.ts
+++ b/src/runtime/query/query.ts
@@ -29,7 +29,7 @@ export const createQuery = <T>(
 
   const query: QueryBuilder<T> = {
     params: () => Object.freeze(params),
-    only: $set('only', ensureArray),
+    only: $set('only', ensureArray) as () => ReturnType<QueryBuilder<T>['only']>,
     without: $set('without', ensureArray),
     where: $set('where', (q: any) => [...ensureArray(params.where), q]),
     sort: $set('sort', (sort: SortOptions) => [...ensureArray(params.sort), ...ensureArray(sort)]),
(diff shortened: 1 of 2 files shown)


Atinux (approved):
Looks good to me! Awesome change. Maybe @farnabaz can have a second review

Tahul:
LGTM :)

Thanks! ❤️

src/runtime/query/query.ts
```diff
@@ -1,9 +1,10 @@
 import type { DatabaseFetcher, QueryBuilder, QueryBuilderParams, SortOptions } from '../types'
+import { ParsedContentMeta } from '../types'
 import { ensureArray } from './match/utils'
 
 const arrayParams = ['sort', 'where', 'only', 'without']
 
-export const createQuery = <T>(
+export const createQuery = <T = ParsedContentMeta>(
```
farnabaz, line 7:
```suggestion
export const createQuery = <T = ParsedContent>(
```

src/runtime/query/query.ts
```diff
@@ -1,9 +1,10 @@
 import type { DatabaseFetcher, QueryBuilder, QueryBuilderParams, SortOptions } from '../types'
+import { ParsedContentMeta } from '../types'
```
farnabaz, line 2:
```suggestion
import { ParsedContent } from '../types'
```

farnabaz (approved):

Tahul (approved):

PR #1208, opened 2022-06-03
fix(navigation): allow navigation opt-out with `navigation: false`

harlan-zw:
### ❓ Type of change

### 📚 Description

The [documentation](https://content.nuxtjs.org/guide/displaying/navigation) says 

> Set navigation: false in the [front-matter](https://content.nuxtjs.org/guide/writing/markdown) of a page to exclude it in fetchContentNavigation() return value.

However, it doesn't work. I couldn't find any logic for excluding this. The Example on the page is even showing the wrong data.

This change properly filters out the pages with this setting from the navigation query.

### 📝 Checklist

### DIFF
--- a/src/runtime/server/api/navigation.ts
+++ b/src/runtime/server/api/navigation.ts
@@ -10,6 +10,14 @@ export default defineEventHandler(async (event) => {
   const contents = await serverQueryContent(event, query)
     .where({
       /**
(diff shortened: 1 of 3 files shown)


Tahul:
Hey 👋,

This is a bad mistake from myself, this behavior is actually implemented into Docus and not Content module 🤦

Thank you so much for fixing that!

Tahul (approved):

PR #1204, opened 2022-06-03
refactor!: spell `extensions` correctly

harlan-zw:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

The key is spelled incorrectly, as someone who is adding a custom transformer I'd prefer not to have misspelled keys in my code. 

BREAKING CHANGE: custom transformers should rename key `extentions` to  `extensions`

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #1213, opened 2022-06-05
fix(transformers): opt-in to correct `extensions` spelling

harlan-zw:
### 🔗 Linked issue

Related https://github.com/nuxt/content/pull/1204#issuecomment-1146647579

### ❓ Type of change

### 📚 Description

This is an alternative solution to the fix for mispelled extensions, this PR reverts the breaking change commit of  https://github.com/nuxt/content/pull/1204#issuecomment-1146647579. We prefer to use a soft breaking change to avoid having to bump majors.

I've also included a test for a custom transformer here using the deprecated key.

### 📝 Checklist

### DIFF
--- a/docs/content/3.guide/2.displaying/4.typescript.md
+++ b/docs/content/3.guide/2.displaying/4.typescript.md
@@ -39,6 +39,27 @@ const { data } = await useAsyncData(
 </script>
 ```
 
+## Markdown Specific Types
+
+If you know the content being fetched will be Markdown, then you can extend the `MarkdownParsedContent`{lang="ts"} type for improved
+type-safety.
+
+```vue
+<script setup lang="ts">
+import type { MarkdownParsedContent } from '@nuxt/content/dist/runtime/types'
+
+interface Article extends MarkdownParsedContent {
+  author: string
+}
+const { data } = await useAsyncData(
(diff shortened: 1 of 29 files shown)


harlan-zw:
Hm I wanted to revert the other commit but have done something wrong 😬

harlan-zw:
Okay should be good

Tahul:
That you a lot @harlan-zw 🙂

Should've renamed the original PR, this is my bad!

The new test is a great addition as well!

Tahul (approved):

farnabaz:
I agree that no one yet used this feature, also we do not have good documentation about parsers and transformers.

Transformers are mainly used internally, We did not expose types and a helper to create custom transformers yet. So I see no harm in dropping misspelled key.

It will be good if we create/expose util to define transformers. like `defineContentTransformer`. This way we ensure users have full ts support and better DX

Tahul (changes requested):
Could you please remove the `@deprecated` keys and support for previous implementation?

I would like to preserve the test-case there, but not the support for older implementations (point about this in discussion with @farnabaz above).

harlan-zw:
I can make a seperate pr for the test case if not needed 👍

PR #1261, opened 2022-06-15
feat(navigation): allow _dir.yml to filter navigation

Tahul:
### ❓ Type of change

- [X] 👌 Enhancement (improving an existing functionality like performance)
- [X] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

The `_dir.yml` has no control over navigation, it only inherits picked fields.

Added the support for `navigation: false` in `_dir.yml`.

I think it should be consistent with Front Matter features.

I also improved the navigation documentation page a lot, and fixed ReadMore component that was missing from docs but used in multiple pages.

### 📝 Checklist


farnabaz (approved):
LGTM 👍

PR #1291, opened 2022-06-27
fix(highlight): warn about languages dynamic loading

farnabaz:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/runtime/server/api/highlight.ts
+++ b/src/runtime/server/api/highlight.ts
@@ -42,8 +42,8 @@ const resolveBody = (body: Partial<HighlightParams>) => {
     // Remove trailing carriage returns
     code: body.code.replace(/\n+$/, ''),
     // Resolve lang & theme (i.e check if shiki supports them)
-    lang: resolveLang(body.lang),
-    theme: resolveTheme(body.theme)
+    lang: resolveLang(body.lang || ''),
+    theme: resolveTheme(body.theme || '')
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/runtime/server/api/highlight.ts
```diff
@@ -87,7 +87,14 @@ export default defineLazyEventHandler(async () => {
 
     // Load supported language on-demand
     if (!highlighter.getLoadedLanguages().includes(lang)) {
-      await highlighter.loadLanguage(lang)
+      // eslint-disable-next-line no-console
+      console.warn(`[Nuxt] [Content] Language "${lang}" is not loaded Shiki. Falling back to plain code.`)
```
Atinux, line 91: Cannot we use our logger here instead? https://github.com/nuxt/content/blob/main/src/utils.ts#L11

```suggestion
      console.warn(`Language "${lang}" is not loaded Shiki. Falling back to plain code.`)
```

src/runtime/server/api/highlight.ts
```diff
@@ -87,7 +87,14 @@ export default defineLazyEventHandler(async () => {
...
+      // eslint-disable-next-line no-console
+      console.warn(`[Nuxt] [Content] Please make sure you add "${lang}" to the 'preload' list in your Nuxt config. See https://content.nuxtjs.org/api/configuration#highlight`)
```
Atinux, line 93: Same here.

Tahul:
image

Here is what the message looks like now

Tahul (approved):

PR #1305, opened 2022-06-29
feat: Unwanted <p> tag Over <img> improved.

belwalpb:
### ❓ Type of change

- [ x] 👌 Enhancement (improving an existing functionality like performance)

### 📚 Description
image
For the Markdown of Above Type Where Image Exists, As Per the Current Implementation the parsed Syntax Tree Would Be:
image
It Means an unwanted p tag was introduced over each **img** tag.
image

**Added The Logic to Check If Current Element is p, It has only one child and that child is img, then unwrap that child.**
image

After the Implementation the Syntax Tree Would Be Like This:
image

### 📝 Checklist

- [ ] I have linked an issue or discussion. [**No Issue was Raised Against It**]
- [ ] I have updated the documentation accordingly. [**Documentation Update Not Required**]

### DIFF
--- a/src/runtime/markdown-parser/compiler.ts
+++ b/src/runtime/markdown-parser/compiler.ts
@@ -20,6 +20,17 @@ export default function (this: any, _options: MarkdownOptions) {
     if (Array.isArray(node)) {
       return node.map(parseAsJSON).filter(Boolean)
     }
+
+
+    /**
+     * If Current Element is p and Its Only Child is img then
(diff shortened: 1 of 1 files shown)


farnabaz:
Hey @belwalpb,
Wrapping `` inside `<p>` is the default behavior of markdown. It is not a good choice to change the default behavior of markdown.

If you want to render images outside of paragraph you can use MDC syntax:

```
::img{src="....."}
::
```

PR #1336, opened 2022-07-06
fix(document-driven): rendering flash

farnabaz:
### ❓ Type of change

### 📚 Description
- Embed NuxtLayout inside the  document-driven page
- Overwrite the main component and remove wrapper NuxtLayout

### 📝 Checklist


src/module.ts
```diff
@@ -448,12 +448,20 @@ export default defineNuxtModule<ModuleOptions>({
...
+              path: '/:slug(.*)*',
+              file: resolveRuntimeModule('./pages/document-driven.vue'),
+              children: []
+            })
+          }
+        })
+        nuxt.hook('app:resolve', (app) => {
+          if (app.mainComponent?.includes('@nuxt/ui-templates')) {
+            app.mainComponent = resolveRuntimeModule('./app.vue')
+          }
```
Atinux, line 464: What do you think of reading the mainComponent when we cannot overwrite it, check if it includes (`<NuxtLayout`) and warn the user to use `<NuxtLayout>` in the page to avoid layout shift with the document driven mode?

Atinux:
Seems we have an issue with HMR when changing the `layout` prop:

https://user-images.githubusercontent.com/904724/177545866-f107a692-896f-4f35-8dc3-95b61ce748d3.mp4

Atinux:
Also seems that the layout is not applied for `http://localhost:3000/vue-file` (`yarn dev document-driven`)

farnabaz:
Added watch in DD page to handle HMR https://github.com/nuxt/content/pull/1336/commits/e63306cb5b4ff72f5aa856c1ba2570aae6ec2ca7

Tahul:
Before merging this could we please link this PR to the issue upstream related to that rendering issue?

Also could we leave TODO's in code pointing to this PR or the tracked issue.

Just want to be sure this is not considered as long term workaround but more as an hot fix we consider acceptable for a minor release, that will transparently be replaced by the long term solution when the upstream issue gets resolved.

farnabaz:
The issue is, when this resolved in upstream, it will be breaking change to users that have custom `app.vue` in their projects

Tahul:
Here is a (incomplete) list of issues to track that will most likely improve the situation:
- https://github.com/vuejs/core/issues/5844
- https://github.com/vuejs/router/issues/1319
- https://github.com/vuejs/router/issues/626

These issues have kindly been forwarded by @danielroe, thanks a lot for this!

These are not directly causing this issue looks more like a combination of factors that could have impact on what is wrong for us.

Considering this PR, I think we can move forward with it and prepare the migration notes for users when they will have to rollback the workarounds we are forced to use to (partially) fix this rendering flash.

Atinux (approved):

PR #1385, opened 2022-07-27
feat: single file source

### LINKED ISSUE #1348: Allow one file sources
We currently only support specifying a directory as a custom source like this:

```
{
   name: 'fa-ir',
   prefix: '/fa', // All contents inside this source will be prefixed with `/fa`
   driver: 'fs',
   base: resolve(__dirname, 'content-fa') // Path for source directory
}
```

Would it be possible to support specifying a complete file path as a source?

It would be useful for cases like:

- Adding `nuxt` and `@nuxt/content` dependencies to a repository
- Only have a `nuxt.config.ts` at the root of the project
- Render a single file like `README.md` with a custom theme as a website for the repository

farnabaz:
### 🔗 Linked issue
resolves #1348 

### ❓ Type of change

### 📝 Checklist


Atinux:
I would rename `fileName` to `filename` since it is accepted for a while now: https://en.wikipedia.org/wiki/Filename

xarthurx:
seems this PR has been a while here.
Is there any updates for the feature? 
I came here as I want to query a single file that contains info all in the frontmatter, and it seems the current **Nuxt Content**'s query framework cannot do that...

PR #1401, opened 2022-08-01
feat: create index for path base search

farnabaz:
### ❓ Type of change

### 📝 Checklist


src/module.ts
```diff
@@ -262,6 +262,11 @@ export default defineNuxtModule<ModuleOptions>({
           route: `/api/${options.base}/query`,
           handler: resolveRuntimeModule('./server/api/query')
         },
+        {
+          method: 'get',
+          route: `/api/${options.base}/find_one/:qid/**:slug`,
```
Atinux, line 267: Could you confirm it works with slug being without any `/` @farnabaz ?

Trying locally on Studio and seems that it sends back to me a 405

Atinux: Might be worth adding tests into radix3 https://github.com/unjs/radix3/blob/main/tests/insert.test.ts

pi0: https://github.com/unjs/radix3/issues/8

farnabaz: @Atinux I'm going to remove this endpoint and integrate Index with `/query` endpoint

PR #1351, opened 2022-07-12
feat(tailwindcss): support tailwindcss classes in content (with hmr)

Tahul:
### 🔗 Linked issue

This PR follows https://github.com/nuxt-community/tailwindcss-module/pull/499 ; and move the support to this module instead based on @pi0's comment.

### ❓ Type of change

- [X] 👌 Enhancement (improving an existing functionality like performance)
- [X] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

Adds first-class support for @nuxtjs/tailwindcss classes usage inside any content file (even remote sources).

Quick demo:
https://twitter.com/yaeeelglx/status/1546643431216369665

I also upgraded @nuxt/content documentation to use latest Docus (and @nuxt-themes/kit)

### 📝 Checklist

- [X] I have updated the documentation accordingly.

---

cc @farnabaz ; how could we test this feature?

PR #1411, opened 2022-08-03
feat: pre fetch contents on build

### LINKED ISSUE #1387: Generate parsed content during module boot in production
Instead of adding the `/api/_content/cache` route in pre-render.

This way, Tailwind JIT (and others) will work.

farnabaz:
### 🔗 Linked issue
resolves #1387

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -17,11 +17,13 @@ import { join } from 'pathe'
 import type { Lang as ShikiLang, Theme as ShikiTheme } from 'shiki-es'
 import { listen } from 'listhen'
 import type { WatchEvent } from 'unstorage'
+import { createStorage } from 'unstorage'
 import { withTrailingSlash } from 'ufo'
 import { name, version } from '../package.json'
 import {
   CACHE_VERSION,
   createWebSocket,
+  getMountDriver,
   logger,
   MOUNT_PREFIX,
   processMarkdownOptions,
@@ -511,8 +513,35 @@ export default defineNuxtModule<ModuleOptions>({
       ...contentContext as any
     }
 
+    // @nuxtjs/tailwindcss support
+    // @ts-ignore - Module might not exist
+    nuxt.hook('tailwindcss:config', (tailwindConfig) => {
+      tailwindConfig.content = tailwindConfig.content ?? []
+      tailwindConfig.content.push(`${nuxt.options.buildDir}/cache/content/parsed/**/*.md`)
+    })
+
     // Setup content dev module
-    if (!nuxt.options.dev) { return }
+    if (!nuxt.options.dev) {
+      nuxt.hook('build:before', async () => {
+        const storage = createStorage()
+        const sources = useContentMounts(nuxt, contentContext.sources)
+        sources.cache = {
+          driver: 'fs',
+          base: resolve(nuxt.options.buildDir, 'cache')
+        }
+        for (const [key, source] of Object.entries(sources)) {
+          storage.mount(key, getMountDriver(source))
+        }
+        const keys = await storage.getKeys()
+        await Promise.all(
+          keys.map(async (key) => {
+            const content = await storage.getItem(key)
+            await storage.setItem(`cache:content:parsed:${key.substring(14)}`, content)
+          })
+        )
+      })
+      return
+    }
 
     nuxt.hook('nitro:init', async (nitro) => {
       if (!options.watch || !options.watch.ws) { return }
@@ -542,13 +571,6 @@ export default defineNuxtModule<ModuleOptions>({
         // Broadcast a message to the server to refresh the page
         ws.broadcast({ event, key })
       })
-    })
-
-    // @nuxtjs/tailwindcss support
-    // @ts-ignore - Module might not exist
-    nuxt.hook('tailwindcss:config', (tailwindConfig) => {
-      tailwindConfig.content = tailwindConfig.content ?? []
-      tailwindConfig.content.push(`${nuxt.options.buildDir}/cache/content/parsed/**/*.md`)
     })
   }
 })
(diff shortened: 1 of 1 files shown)


src/module.ts
```diff
@@ -511,8 +513,50 @@ export default defineNuxtModule<ModuleOptions>({
       ...contentContext as any
     }
 
+    // @nuxtjs/tailwindcss support
+    // @ts-ignore - Module might not exist
+    nuxt.hook('tailwindcss:config', (tailwindConfig) => {
+      tailwindConfig.content = tailwindConfig.content ?? []
+      tailwindConfig.content.push(`${nuxt.options.buildDir}/cache/content/parsed/**/*.md`)
```
pi0, line 520: Important note: This path is highly liked to be changed for `.nuxt/cache` to differentiate between dev and prod.

src/module.ts
```diff
@@ -511,8 +513,50 @@ export default defineNuxtModule<ModuleOptions>({
...
+
     // Setup content dev module
-    if (!nuxt.options.dev) { return }
+    if (!nuxt.options.dev) {
+      nuxt.hook('build:before', async () => {
+        const storage = createStorage()
+        const sources = useContentMounts(nuxt, contentContext.sources)
+        sources.cache = {
+          driver: 'fs',
+          base: resolve(nuxt.options.buildDir, 'cache')
```
pi0, line 530: Please use module prefix for custom cache dir (can be `.nuxt/content-cache`)

src/module.ts
pi0: Design note: This integration could be in tailwind. It is tailwind that being extended by content not content by tailwind.

src/module.ts
farnabaz: This is not a custom cache, this code tries to initialize and fill Nuxt cache. Having another directory for the cache will result duplicating contents and tasks

pi0: You can mount `/cache/content` to `/cache` this way you can both predict real fs path and also any write to `/cache/content` goes there not being affected by nuxt and nitro defaults.

src/module.ts
farnabaz: Is there any util or option to get cache dir?

pi0: Check https://github.com/nuxt/content/pull/1411#discussion_r940226190 for better solution Generally never depend on buildDir filesystem structure.

farnabaz: > Design note: This integration could be in tailwind. It is tailwind that being extended by content not content by tailwind.

This integration moved to the content module (#1351) based on a discussion in the tailwind module.  https://github.com/nuxt-community/tailwindcss-module/pull/499#discussion_r918823965

farnabaz: > Check https://github.com/nuxt/content/pull/1411#discussion_r940226190 for better solution Generally never depend on buildDir filesystem structure.

This implies double fetching remote sources even in dev mode, which had performance downsides for users in dev mode.

pi0: > This implies double fetching remote sources even in dev mode, which had performance downsides for users in dev mode.

No it won't! You just readirect fs writes :)

> Design note: This integration could be in tailwind. It is tailwind that being extended by content not content by tailwind. (resolving to discuss in other comment)

> This integration moved to the content module (https://github.com/nuxt/content/pull/1351) based on a discussion in the tailwind module. https://github.com/nuxt-community/tailwindcss-module/pull/499#discussion_r918823965

Ah my bad. Let's do this later.

src/module.ts
pi0: > This implies double fetching remote sources even in dev mode, which had performance downsides for users in dev mode.

@farnabaz Why do you think it cases double fetching? By default keys to `/cache/*` are mapped to `.nuxt/cache/*`. By mounting, it remains same and any write to `/cache/content/*` maps to `/.nuxt/content-cache/*` instead).

farnabaz: Right now we are fetching remote sources twice, once in `before:build` hook and another in nitro runtime. But this only happens in build time. If we change tailwind search dir, we need to fetch contents in development mode too. And have duplicate fetch in dev mode.

pi0: Fetching would be the same also for dev as unstorage keys remain the same as before. Maybe i can explain better this way, currently we assume during build and dev, a key write to `cache:content:xyz` maps to `.nuxt/cache/content/xyz` file. This might change in the future. Nuxt might use `.nuxt/cache/[dev|prod]/content/xyz` or even `.nuxt/cache/db.sql`. 

If you (dev-build and prod-build) mount a nitro storage to `/cache/content` and explicitly mapping to single fs path `.nuxt/content-cache`, you can be sure a write to `cache:content:xyz` goes to certain directory structure expected by tailwind or any other internal wants to interact with cache without storage layer but directly accessing it.

With what i suggested, both tailwind and content, still write to **one path** but a predictable one that you expect :)

This change is probably not directly related to this PR but since i spotted a hard code dependency on fs path, noted this to be solved sooner and avoid breaking content.

farnabaz: Are you suggesting adding `/cache/content` mount to nitro storage?

farnabaz: @pi0  by `.nuxt/content-cache`, you mean `${nuxt.option.rootDir}/.nuxt` ? or keep using `buildDir`?

pi0: Yes. I use `.nuxt` instead of `buildDir` in examples to be readable :)

src/module.ts
```diff
@@ -278,6 +278,10 @@ export default defineNuxtModule<ModuleOptions>({
       // Register source storages
       const sources = useContentMounts(nuxt, contentContext.sources)
       nitroConfig.devStorage = Object.assign(nitroConfig.devStorage || {}, sources)
+      nitroConfig.devStorage['cache:content'] = {
+        driver: 'fs',
+        base: resolve(nuxt.options.rootDir, '.nuxt', 'content-cache')
```
pi0, line 283: `resolve(buildDir, 'content-cache')` was fine. Sorry i only used as example with `.nuxt` @farnabaz

PR #1418, opened 2022-08-05
fix: make sure `components/content` is on top in layers

CodeDredd:
### 🔗 Linked issue

closes #1417, related to nuxt/framework#6382

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

In mutli layer setups the priority of the layers is wrong and needs to be reversed, so that `content` is on top

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #1404, opened 2022-08-02
fix: support components/content in layers for extends

Atinux:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Resolves #1386 

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #1455, opened 2022-08-16
fix: avoid mutating `_layers`

pi0:
### ❓ Type of change

### 📚 Description

We introduced a fix in #1418 to reverse scan layers and regression of #1404. By calling `reverse()` we were mutating original `_layers` too causing whole layer scanning logic after content being broken.

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -356,7 +356,8 @@ export default defineNuxtModule<ModuleOptions>({
     })
 
     // Register user global components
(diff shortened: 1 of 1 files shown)


pi0:
An alternative would be not reversing layers and instead `push` component dirs (we can do in another PR as currently scanning is broken).

/cc @Atinux Do you remember why had to originally unshift component dirs instead of pushing?

farnabaz (approved):

PR #1496, opened 2022-09-05
feat(config): add toggle for `prefetch` links

harlan-zw:
### 🔗 Linked issue

Related: https://github.com/danielroe/nuxt-zero-js/pull/8

### ❓ Type of change

### 📚 Description

When using [nuxt-zero-js](https://github.com/danielroe/nuxt-zero-js/) or if you've setup the content components to be server rendered only, you'll be in a situation where you don't need the API data pre-fetched.

Currently, there is no way to remove the pre-fetches without regex matching on the html output.

I've intentionally avoided adding doc as this is a config option that either advanced users or module authors would take advantage of.

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -182,7 +182,13 @@ export interface ModuleOptions {
     }
     layoutFallbacks: string[]
     injectPage: boolean
-  }
+  },
+  /**
+   * Toggle meta `prefetch` tags being added to the head when making fetch requests.
+   *
+   * @default true
+   */
+  prefetch: boolean
 }
 
 interface ContentContext extends ModuleOptions {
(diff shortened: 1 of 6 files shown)


Atinux:
Thanks @harlan-zw 

Actually we shall update to not use useHead() but use headers hints to tell Nitro to pre-render those API routes directly.

This will also fix your issue.

Can you confirm @pi0 ?

pi0:
I think it makes sense to only use nitro prerender header 👍🏼  For performance of loading pages, result of API should already be in payload so prefetching doesn't makes sense.

harlan-zw:
Sounds good, thanks for the prompt reply :+1: 

Will close this off

PR #1501, opened 2022-09-05
feat: create chunk for components

farnabaz:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -348,6 +348,17 @@ export default defineNuxtModule<ModuleOptions>({
       global: true
     })
 
+    // Use single components chunk
+    nuxt.hook('vite:extendConfig', (config, { isServer }) => {
+      if (Array.isArray(config.build.rollupOptions.output) || isServer) { return }
+      const moduleComponentsDir = resolve('./runtime/components')
+      config.build.rollupOptions.output.manualChunks = (id) => {
+        if (id.includes(moduleComponentsDir)) {
+          return 'content-components'
+        }
+      }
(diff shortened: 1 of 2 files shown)


src/module.ts
```diff
@@ -348,6 +348,17 @@ export default defineNuxtModule<ModuleOptions>({
       global: true
     })
 
+    // Use single components chunk
+    nuxt.hook('vite:extendConfig', (config, { isServer }) => {
+      if (Array.isArray(config.build.rollupOptions.output) || isServer) { return }
+      const moduleComponentsDir = resolve('./runtime/components')
+      config.build.rollupOptions.output.manualChunks = (id) => {
```
pi0, line 355: Would you please add a condition to not reassign `manualChunks` if already set? By implementing https://github.com/nuxt/framework/issues/7240 it will cause conflicting behavior.

Atinux:
I believe that could be nice to do the same for global components as well in `components/content/` (supporting extends too)

farnabaz:
> I believe that could be nice to do the same for global components as well in components/content/ (supporting extends too)

Indeed, I think it will be good to generate a separate chunk for them.


Note!!! Currently, the `vue-router` and some other packages are bundling into our custom chunk instead of entry. The chunk size is around 120kb with Vue router and others.

Atinux:
Good luck finding this 😅

dargmuesli:
Is there a ticket or description on what this change improves? I'm interested! :nerd_face:

PR #1545, opened 2022-09-20
feat(document-driven): support `navigation.redirect` from `_dir` files

farnabaz:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/runtime/plugins/documentDriven.ts
+++ b/src/runtime/plugins/documentDriven.ts
@@ -1,7 +1,7 @@
 import type { RouteLocationNormalized, RouteLocationNormalizedLoaded } from 'vue-router'
 // @ts-ignore
-import { useRuntimeConfig, addRouteMiddleware } from '#app'
-import { withoutTrailingSlash } from 'ufo'
+import { useRuntimeConfig, addRouteMiddleware, callWithNuxt, navigateTo } from '#app'
+import { withoutTrailingSlash, hasProtocol } from 'ufo'
 import { NavItem, ParsedContent } from '../types'
 // @ts-ignore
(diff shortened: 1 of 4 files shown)


src/runtime/plugins/documentDriven.ts
```diff
@@ -224,6 +224,7 @@ export default defineNuxtPlugin((nuxt) => {
 
       // Use `redirect` key to redirect to another page
       if (_page?.redirect) { return _page?.redirect }
+      if (_page?._dir?.navigation?.redirect) { return _page?._dir?.navigation?.redirect }
```
Atinux, line 227: I think we can leverage `navKeyFromPath(_page._path, 'redirect', navigation)` instead of fetching the `_dir.yml`

src/runtime/server/api/query.ts
```diff
@@ -1,11 +1,26 @@
...
 
 export default defineEventHandler(async (event) => {
   const query = getContentQuery(event)
   const contents = await serverQueryContent(event, query).find()
 
+  if (query.first) {
+    const path = contents?._path || query.where.find(w => w._path)?._path
+    if (path) {
+      const _dir = await serverQueryContent(event).where({ _path: join(path, '_dir') }).without('_').findOne()
+      if (!Array.isArray(_dir)) {
```
Atinux, line 14: Why checking if it is not an array? In which case `findOne()` returns an Array?

farnabaz: It returns an empty array only when there is no matched document.
We will fix this behavior in another PR

Atinux (approved):

PR #1556, opened 2022-09-24
fix: use relative `.md` link

kecrily:
### ❓ Type of change

### 📚 Description

Before:

`[Test](./1.test.md)` -> `/1.test.md`

After:

`[Test](./1.test.md)` -> `/test`

### 📝 Checklist

### DIFF
--- a/src/runtime/markdown-parser/handler/link.ts
+++ b/src/runtime/markdown-parser/handler/link.ts
@@ -28,8 +28,8 @@ export default function link (h: H, node: Node) {
 }
 
 function normalizeLink (link: string) {
-  if (isRelative(link)) {
-    return link.replace(/^[0-9]*\./g, '').replace(/\.md$/g, '')
+  if (isRelative(link) || (!/^https?/.test(link) && !link.startsWith('/'))) {
+    return link.replace(/^(..?\/)?[0-9]*\./g, '$1').replace(/\.md$/g, '')
   } else {
     return link
   }
--- a/src/runtime/markdown-parser/handler/link.ts
+++ b/src/runtime/markdown-parser/handler/link.ts
@@ -30,10 +30,9 @@ export default function link (h: H, node: Node) {
 function normalizeLink (link: string) {
   if (isRelative(link) || (!/^https?/.test(link) && !link.startsWith('/'))) {
     return link.split('/')
-      .map(x => x
-        .replace(/^[0-9]*\./g, '')
-        .replace(/\.md$/g, ''))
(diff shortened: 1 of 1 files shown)


src/runtime/markdown-parser/handler/link.ts
```diff
@@ -25,3 +26,11 @@ export default function link (h: H, node: Node) {
 
   return h(node, 'a', props, all(h, node))
 }
+
+function normalizeLink (link: string) {
+  if (isRelative(link)) {
+    return link.replace(/^[0-9]*\./g, '').replace(/\.md$/g, '')
```
farnabaz, line 32:
```suggestion
    return link.replace(/^(.\/)?[0-9]*\./g, '$1').replace(/\.md$/g, '')
```

src/runtime/markdown-parser/handler/link.ts
```diff
@@ -25,3 +26,11 @@ export default function link (h: H, node: Node) {
 
   return h(node, 'a', props, all(h, node))
 }
+
+function normalizeLink (link: string) {
+  if (isRelative(link)) {
```
farnabaz, line 31: `isRelative` only detects links starting with `./`, `../`. It will be good if we support links without these prefixes.

```
[test](1.test.md)
```

src/runtime/markdown-parser/handler/link.ts
kecrily: It forgets `../` situation.

farnabaz: True, we should also think about multipart links

`./1.test/2.index.md`

src/runtime/markdown-parser/handler/link.ts
```diff
@@ -25,3 +26,15 @@ export default function link (h: H, node: Node) {
...
   return h(node, 'a', props, all(h, node))
 }
+
+function normalizeLink (link: string) {
+  if (isRelative(link) || (!/^https?/.test(link) && !link.startsWith('/'))) {
+    return link.split('/')
+      .map(x => x
+        .replace(/^[0-9]*\./g, '')
+        .replace(/\.md$/g, ''))
+      .join('/')
```
farnabaz, line 36:
```suggestion
      .map(x => x.replace(/^[0-9]*\./g, ''))
      .join('/')
      .replace(/\.md$/g, '')
```

We only need to remove `.md` from the file name

farnabaz (approved):
Thanks

PR #1564, opened 2022-09-28
feat: config for anchor link generation of headings

### LINKED ISSUE #1444: Default Prose Headings components should not render as links (bad HTML and too opinionated)

As you can see here [main/src/runtime/components/Prose/ProseH3.vue](https://github.com/nuxt/content/blob/1593ad72d37ee870b348d8d66852223d551bd123/src/runtime/components/Prose/ProseH3.vue), H2 to H6 headings become links. 

- It's not good to have such opiniated structure for something so basic like Heading HTML elements. 

- The V1 option was better, i.e., having hidden `#` on left side of heading which shows on `hover`, instead of converting entire heading text into `a`.

- Also, it's not good HTML too, because such `a` link is [inline](https://developer.mozilla.org/en-US/docs/Web/HTML/Element#inline_text_semantics). And `a` link cannot have [certain descendants](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#properties). But currently, the component accepts any slot, so any random HTML element will be descendant of `a`.

- Also, it makes it harder to select the Text of the Heading. 

Just add simple `slot` like https://github.com/nuxt/content/blob/main/src/runtime/components/Prose/ProseH1.vue.  
Nothing fancy. 

For those who want links in their headings, they can add it using respective remark plugin (e.g., autolink-headings). That plugin gives way more options any way. 

Create own Prose components for H2 to H6 headings. 👎🏻 Because it's like resetting something, whereas styling and how template should be --- such things any way should be left to the user, and not Module.

### Additional context

Tailwind's [Preflight philosophy](https://tailwindcss.com/docs/preflight). As a Module, have sensible defaults. Don't have such opiniated rigidity that user have to spend time resetting things anyway. 

> Simply make things behave more like you’d expect them to.

nobkd:
### 🔗 Linked issue
#1444 

### ❓ Type of change

### 📚 Description

Config can set if and for which headings to generate an anchor link
Resolves #1444 

Default is set to generate anchor links for h2 to h4 like before

### 📝 Checklist

### TODO
- [x] Move anchorLinks option to Markdown field
- [ ] Test updated code
- [ ] Update docs


src/module.ts
```diff
@@ -186,6 +186,25 @@ export interface ModuleOptions {
     layoutFallbacks?: string[]
     injectPage?: boolean
   },
+  /**
+   * Anchor link generation config
+   *
+   * @default {}
+   */
+  anchorLinks?: boolean | {
```
farnabaz, line 194: Shall we move this option under the `markdown` field?

nobkd: Sounds reasonable. We should do that.

farnabaz (approved):
Thanks ❤️

PR #1643, opened 2022-10-30
docs: fix error in ContentQuery where clause example

BrettOrmsby:
### ❓ Type of change

### 📚 Description

Fixed an error in the Where clause for ContentQuery documentation. Before, ContentQuery passed `data` variable to the slot, which was not used and instead `authors` was used in its place. `data` is replaced with `authors` to fix this error.

### 📝 Checklist

### DIFF
--- a/docs/content/4.api/1.components/4.content-query.md
+++ b/docs/content/4.api/1.components/4.content-query.md
@@ -62,9 +62,9 @@ The `not-found`{lang=ts} slot can be used to display a default content before re
 ```html [pages/about.vue]
 <template>
   <main>
-    <ContentQuery path="/about/authors" :where="{ type: 'csv' }" v-slot="{ authors }">
+    <ContentQuery path="/about/authors" :where="{ type: 'csv' }" v-slot="{ data }">
       <ul>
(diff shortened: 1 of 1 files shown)


docs/content/4.api/1.components/4.content-query.md
```diff
@@ -62,7 +62,7 @@ The `not-found`{lang=ts} slot can be used to display a default content before re
 ```html [pages/about.vue]
 <template>
   <main>
-    <ContentQuery path="/about/authors" :where="{ type: 'csv' }" v-slot="{ data }">
+    <ContentQuery path="/about/authors" :where="{ type: 'csv' }" v-slot="{ authors }">
```
farnabaz, line 65: We should keep `data` here, and change it in the `v-for` below

```
        <li v-for="author of data" :key="author.name">
          {{ author.name }}
        </li>
```

BrettOrmsby: Yes, my mistake. It is fixed now.

farnabaz (approved):

PR #1752, opened 2022-12-14
feat(query): fully cacheable api

farnabaz:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/runtime/composables/query.ts
+++ b/src/runtime/composables/query.ts
@@ -1,9 +1,9 @@
 import { joinURL, withLeadingSlash, withoutTrailingSlash } from 'ufo'
 import { hash } from 'ohash'
-import { useCookie, useRuntimeConfig } from '#app'
+import { useRuntimeConfig } from '#app'
 import { createQuery } from '../query/query'
 import type { ParsedContent, QueryBuilder, QueryBuilderParams } from '../types'
-import { jsonStringify } from '../utils/json'
+import { encodeQueryParams } from '../utils/query'
 import { addPrerenderPath, shouldUseClientDB, withContentBase } from './utils'
 
 /**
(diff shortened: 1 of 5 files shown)


farnabaz:
Merging to test on edge

src/runtime/composables/query.ts
```diff
@@ -37,7 +37,7 @@ export const createQueryFetch = <T = ParsedContent>(path?: string) => async (que
 
   const params = query.params()
 
-  const apiPath = withContentBase(process.dev ? '/query' : `/query/${hash(params)}.${content.integrity}.json`)
+  const apiPath = withContentBase(`/query/${process.dev ? '_' : `${hash(params)}.${content.integrity}`}/${encodeQueryParams(params)}.json`)
```
pi0, line 40: Important note: This is potentially a breaking change as Nitro v1 limits whole URL length to 250 and by this change, we cause to skip such routes when prerendering. 

https://github.com/unjs/nitro/blob/cbcd0681d37839f93fc7f3f106d34e3f01d2ac84/src/prerender.ts#L83

This change could be behind an experimental flag until the release of Nuxt 3.1 and Nitro 2

pi0: Also, this change causes slight performance overhead of long URLs which is mainly needed for hybrid/server mode. When `_generate: true` flag exists, we can safe assume deployment is full-static and omit last segment.

pi0: https://github.com/unjs/nitro/pull/757

farnabaz: So what you are suggesting is that we should switch between these methods based on `nuxt.options._generate` boolean.
Pass this option to the runtimeconfig to delect whether its generated or not and change the API fetch.

Right?
@pi0

farnabaz: Personally, I don't like to create complicated behavior for us to describe and for users to understand.
So I'll go with experimental flag and disable this feature by default.

pi0: That should work yes. We also need to have it as an experimental opt-in flag as mentioned above, this feature, needs to be coupled with nitro v2 fix for long path segments support.

PR #1800, opened 2023-01-06
feat(markdown): keep meta from fenced code block

Qwertovsky:
### ❓ Type of change

### 📚 Description

Don't drop a info from fenced code
````markdown
```javascript[file.js]{4-6,7} meta-info=val
export default () => {
  console.log('Code block')
}
```
````
So `meta-info=val` will be accessible from the document tree after parsing. Property `meta` will be next to `language`, `filename`, `highlights`.
It can help to do other modifications with remark plugins, transformers, or ProseCode component.

### 📝 Checklist

### DIFF
--- a/test/features/parser-markdown.ts
+++ b/test/features/parser-markdown.ts
@@ -45,10 +45,10 @@ export const testMarkdownParser = () => {
         body: {
           id: 'content:index.md',
           content: [
-          '``` ts [file.ts]{4-6,7} other code block info',
-          'let code = undefined;',
-          'return code;',
-          '```'
+            '```ts [file.ts]{4-6,7} other code block info',
+            'let code = undefined;',
+            'return code;',
+            '```'
           ].join('\n')
         }
       })
--- a/test/features/parser-markdown.ts
+++ b/test/features/parser-markdown.ts
@@ -39,6 +39,31 @@ export const testMarkdownParser = () => {
       expect(parsed.body).toHaveProperty('children[0].children[0].tag', 'code-inline')
     })
 
+    test('Keep meta from fenced code block', async () => {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by farnabaz — APPROVED** [review]
LGTM 👍

**Comment by farnabaz** [comment]
@Qwertovsky do you mind fixing linter issues?

File: test/features/parser-markdown.ts [hunk 1]
```diff
@@ -39,6 +39,31 @@ export const testMarkdownParser = () => {
...
+    test('Keep meta from fenced code block', async () => {
+      const parsed = await $fetch('/api/parse', {
+        method: 'POST',
+        body: {
+          id: 'content:index.md',
+          content: [
+          '``` ts [file.ts]{4-6,7} other code block info',
+          'let code = undefined;',
+          'return code;',
+          '```'
```
nobkd, line 51:
```suggestion
            '```ts [file.ts]{4-6,7} other code block info',
            'let code = undefined;',
            'return code;',
            '```'
```

nobkd (changes requested):

SimonScholz:
The sample mentioned above using `{4-6,7}` is not targeting the sample code since it has only 3 lines.

image

This is really unfortunate because I am currently facing the issue that no different highlighting is applied to the lines in brackets, e.g., `{4-6,7}` 
Shall I open a bug for this or is the `{4-6,7}` somehow not supported?

nobkd:
@SimonScholz
Thos should be fixed with nuxt-modules/mdc#21 and will be fixed when the package is updated in Content. If that's the case, then the edge channel should already have the fix. Otherwise it should be fixed when a new version of Content gets released

PR #1801, opened 2023-01-06
feat(markdown): don't create excerpt if there is no <!--more-->

Qwertovsky:
### ❓ Type of change

### 📚 Description
If there is no `<!-- more -->` divider in the text then the excerpt is a copy of full content.

This change adds new boolean variable `excerptIsBody` into the document next to `excerpt`.

- It can help to improve performance and don't process the excerpt (just make a copy of body).
- It can help to don't show link to full text if there is nothing to read more. 

### 📝 Checklist

### DIFF
--- a/docs/content/3.guide/1.writing/2.markdown.md
+++ b/docs/content/3.guide/1.writing/2.markdown.md
@@ -87,6 +87,8 @@ Full amount of content beyond the more divider.
 
 Description property will contain the excerpt content unless defined within the Front Matter props.
 
+If there is no `<!--more-->` divider in the text then excerpt is undefined.
+
 Example variables will be injected into the document:
 
 ```json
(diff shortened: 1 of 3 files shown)


Atinux:
What about making the excerpt empty instead to avoid double payload?

farnabaz (changes requested):
Instead of having an extra field in the document, we should clear the `excerpt` value.

It also reduces the document sizes and smaller chunks.

Qwertovsky:
I made changes. Now the excerpt is undefined.

By default, `<ContentRenderer>`  still shows body.
```ts
    // Get body from value
    let body = (value.body || value) as MarkdownNode
    if (ctx.excerpt && value.excerpt) {
      body = value.excerpt
    }
```
But this PR could be breaking change for some cases. If the excerpt is used as content property in code.

farnabaz (approved):

PR #1818, opened 2023-01-11
fix: sanitize highlighted text

danielroe:
### ❓ Type of change

### 📚 Description

Previously we were producing code like this:

```html
<span class=\"ct-2c497b\"><</span>
```

which breaks the dom tree - or worse, renders the highlighted code directly as HTML elements.

Looking through the codebase it seems likely that this fix might need to be applied elsewhere too and it would be worth a quick look.

### 📝 Checklist


farnabaz (approved):
Thanks 👍

Issue #1823, opened 2023-01-14
import type assertion when defining custom transformers


SheetJSDev:
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.13.0`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `unknown`
- Builder:          `vite`
- User Config:      `modules`, `content`
- Runtime Modules:  `normalizedModule()`, `@nuxt/content@2.3.0`
- Build Modules:    `-`

### Reproduction

https://github.com/nuxt/content/tree/main/examples/advanced/transformer (literally the demo in the source tree)

### Describe the bug

```
 ERROR  [worker reload] [worker init] Module "file:///path/to/nuxt/content/examples/advanced/transformer/node_modules/@nuxt/content/dist/runtime/transformers/shiki/languages/mdc.tmLanguage.json" needs an import assertion of type "json"
```

SheetJSDev:
This was fixed by #1824 .  Until the next version is published (current is `v2.3.0`), the workaround is:

```ts
import { defineTransformer } from '@nuxt/content/transformers/utils'
```

PR #1824, opened 2023-01-14
fix: JSON import type assertion

SheetJSDev:
### 🔗 Linked issue

https://github.com/nuxt/content/issues/1823

### ❓ Type of change

### 📚 Description

See linked issue

### 📝 Checklist

- [ ] I have updated the documentation accordingly. (n/a)

### DIFF
--- a/src/runtime/transformers/shiki/languages/mdc.tmLanguage.ts
+++ b/src/runtime/transformers/shiki/languages/mdc.tmLanguage.ts
@@ -1,3 +1,4 @@
+// @ts-nocheck
 // see microsoft/vscode-textmate#185
 // import { IRawGrammar } from 'vscode-textmate/release/rawGrammar';
 
--- a/src/runtime/transformers/shiki/languages/mdc.tmLanguage.ts
+++ b/src/runtime/transformers/shiki/languages/mdc.tmLanguage.ts
@@ -1,4 +1,7 @@
-{
+// see microsoft/vscode-textmate#185
+// import { IRawGrammar } from 'vscode-textmate/release/rawGrammar';
+
+const grammar /*: IRawGrammar */ = {
   "information_for_contributors": [
     "This file has been converted from https://github.com/docusgen/vscode-extension/blob/main/syntaxes/mdc.tmLanguage.json",
     "If you want to provide a fix or improvement, please create a pull request against the original repository.",
@@ -64,7 +67,7 @@
           "include": "#markdown_attributes"
         }
       ]
-    }, 
+    },
     "markdown_attributes": {
       "match": "(?x)([^ ])(               # attributes\n    ({)\n      ([^{]*)\n    (})\n  )",
       "name": "markup.component.attribute",
@@ -571,3 +574,5 @@
     }
   }
 }
+
+export default grammar;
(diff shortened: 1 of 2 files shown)


HigherOrderLogic:
```cmd
 WARN  rollup-plugin-inject: failed to parse /node_modules/@nuxt/content/dist/runtime/transformers/shiki/highlighter.mjs. Consider restricting the plugin to particular files via options.include           

 ERROR                                                                                                                                                                                                           

2: import consola from "consola";
3: import { createSingleton } from "../utils.mjs";
4: import mdcTMLanguage from "./languages/mdc.tmLanguage.json" assert { type: "json" };
                                                               ^
5: const logger = consola.withScope("@nuxt/content");
6: const resolveLang = (lang) => BUNDLED_LANGUAGES.find((l) => l.id === lang || l.aliases?.includes(lang))?.id || lang;
```

Do you get this error?

SheetJSDev:
That error shows up but oddly enough the pages are still built locally.  This patch was the obvious one to resolve the esbuild issue.

There is a bypass without patching the library: change the `defineTransformer` import to use the version in `utils`.  This change would be applied to the transformer script:

```diff
-import { defineTransformer } from '@nuxt/content/transformers'
+import { defineTransformer } from '@nuxt/content/transformers/utils'
```

HigherOrderLogic:
> That error shows up but oddly enough the pages are still built locally. This patch was the obvious one to resolve the esbuild issue.
> 
> There is a bypass without patching the library: change the `defineTransformer` import to use the version in `utils`. This change would be applied to the transformer script:
> 
> ```diff
> -import { defineTransformer } from '@nuxt/content/transformers'
> +import { defineTransformer } from '@nuxt/content/transformers/utils'
> ```

Thanks for the workaround.

farnabaz:
Import assertion is only supported in Node.js 17 or later ([ref](https://github.com/nodejs/node/commit/95e4d29eb4)). And most of the users still use v16.

farnabaz:
Let's convert the json file into a JavaScript one. This will resolve the issue.

@pi0 WDYT?

SheetJSDev:
That approach is better.  Patch updated.

There were type issues which prevented the use of the official `IRawGrammar` type, mainly around comments.  For example:

```js
{
  "repository": {
    "block": {
      "comment": "Same as `text.html.markdown#block`, but without `raw_block`",
```

Based on its usage:
- the root object should be `IRawGrammar`
- `repository` should be an object of type `IRawRepository`
- `block` should be an object of type `IRawRule`. 

Neither `IRawRule` nor its parent type `ILocatable` have a comment field.  It's unclear whether this oversight is a bug in `vscode-textmate` or if it should be removed here.

src/runtime/transformers/shiki/languages/mdc.tmLanguage.ts
```diff
@@ -1,4 +1,7 @@
-{
+// see microsoft/vscode-textmate#185
```
farnabaz, line 1:
```suggestion
// @ts-nocheck
// see microsoft/vscode-textmate#185
```

src/runtime/transformers/shiki/languages/mdc.tmLanguage.ts
```diff
@@ -1,4 +1,8 @@
-{
+// @ts-nocheck
```
farnabaz, line 1:
```suggestion
/* eslint-disable */
```

farnabaz (approved):
Thanks

PR #1839, opened 2023-01-22
feat: add modified time

### LINKED ISSUE #1797: Missing `createdAt` and `updatedAt` field
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v16.14.0`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `yarn@1.22.19`
- Builder:          `vite`
- User Config:      `srcDir`, `modules`, `colorMode`, `postcss`, `vite`
- Runtime Modules:  `@nuxtjs/tailwindcss@6.2.0`, `@nuxtjs/color-mode@3.2.0`, `@nuxt/content@2.3.0`
- Build Modules:    `-`
------------------------------

### Reproduction

Coming

### Describe the bug

I want to display the createAt from my blog post and sort my posts by date creation date, but I don't see the `createdAt` or `updatedAt` fields in my data unlike what the documentation website says [here](https://content.nuxtjs.org/v1/getting-started/writing#:~:text=The%20createdAt%20and%20updatedAt%20properties%20are%20based%20on%20the%20file%27s%20actual%20created%20%26%20updated%20datetime%2C%20but%20you%20can%20override%20them%20by%20defining%20your%20own%20createdAt%20and%20updatedAt%20values.%20This%20is%20especially%20useful%20if%20you%20are%20migrating%20your%20past%20blog%20posts%20where%20the%20createdAt%20can%20be%20months%20or%20years%20ago.).
Capture d’écran, le 2023-01-05 à 22 53 50

### Additional context

_No response_

### Logs

_No response_

Barbapapazes:
### ❓ Type of change

### 📝 Checklist


Barbapapazes:
Can use https://github.com/unjs/unstorage/issues/136

Barbapapazes:
Unfortunately, the created time can't be get from unstorage because it create a new file and return time from this new file.

Barbapapazes:
After deeper investigations, it seems that `unstorage` won't help to solve this issue (adding the created at time). (`unstorage` will only help for the `updatedAt` meta. But because using this method, we can't get `createdAt`, I think it's not the best option.)

However, in the file `src/runtime/markdown-parser/index.ts`, at the line 57, there is a function called `parseFrontMatter` which return `content` and `data`. And because the `parse` method return a `meta` object, I think we could use it to get `updatedAt`and `createdAt`.
Unfortunaly, this could be a solution but `git` does not store timestamps. (https://stackoverflow.com/questions/62039244/does-git-store-the-file-creation-time).

So, there is no way to easily get creation date from a content file.

Benbinbin:
I also want to figure out how to get the created and updated time.

Check out the starckoverflow you mention above, the [answer](https://stackoverflow.com/a/62039285/10699431) provide an idea: each commit has timestamp

So my solution is searching the added commit and last update commit associated with the specify file and get the timestamp

I write [a server plugin](https://github.com/Benbinbin/BlogiNote/blob/main/server/plugins/gitTimestamp.ts) for my Nuxt Content project [BlogiNote](https://github.com/Benbinbin/BlogiNote)

```ts
/**
 * add timestamp for markdown file
 */
import { execSync } from 'node:child_process'
// ...

export default defineNitroPlugin((nitroApp) => {
   // ...
    const firstCommitTime = execSync(`git log --diff-filter=A --format=%as -- ${filePath}`, { encoding: 'utf-8' }).trim();

   // ...
    const lastCommitTime = execSync(`git log -1 --format=%as -- ${filePath}`, { encoding: 'utf-8' }).trim();

 // ...
})
```

This won't be the precise time as the created and updated, and it works!

However I encouter another problem, this method just work fine locally. 

When I deploy to Vercel, because [the `.git` directory is not deployed](https://github.com/vercel/vercel/discussions/4101), so it fail to get the commit time from git log

PR #1867, opened 2023-01-30
chore(deps): drop html-tags

Tahul:
### ❓ Type of change

- [X] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [X] 👌 Enhancement (improving an existing functionality like performance)

### 📚 Description

Drop `html-tags` dependency as it was breaking builds in some contexts.

As it is a very simple array of HTML tags, I think we can inline it as we do for PROSE_TAGS.

### 📝 Checklist


package.json
```diff
@@ -40,6 +40,7 @@
     "test:coverage": "vitest --coverage",
     "test:types": "tsc --build tsconfig.json",
     "test": "./scripts/test.sh",
+    "prepare": "nuxi prepare playground/basic",
```
farnabaz, line 43:
```suggestion
```

We have `dev:prepare` in scripts

src/runtime/utils/html-tags.ts
```diff
@@ -0,0 +1,119 @@
+export default [
```
farnabaz, line 1:
```suggestion
// Ref https://github.com/sindresorhus/html-tags/blob/v3.2.0/html-tags.json
export default [
```

Do you mind also fixing lints?

Tahul: Sure

package.json
Tahul: Moved `dev:prepare` into `prepare` so it runs at the end of install script :)

farnabaz (approved):

Issue #1666, opened 2022-11-08
Locale detection


ineshbose:
### Is your feature request related to a problem? Please describe

I feel that getting locale (even though the i18n module isn't integrated yet) could be switched.
For eg, I have an About page in English and Spanish (About -> Sobre). If I have `content/en/about.md` and `content/es/sobre.md`, this makes navigating through the codebase a bit difficult while I'm preferring to have `content/about/{en,es}.md`

https://github.com/nuxt/content/blob/main/src/runtime/transformers/path-meta.ts#L33

### Describe the solution you'd like

Option to configure the behaviour of the transformer possibly or something?

### Describe alternatives you've considered

Have two files as `content/en/about.md` and `content/es/sobre.md`. Not preferring to do this.

ineshbose:
Putting a permalink in case the file changes.

https://github.com/nuxt/content/blob/8408d15d5132bdba3cbd62b346a1d5174b049b7a/src/runtime/transformers/path-meta.ts#L33

Maybe a function can be passed in the options?

PR #1899, opened 2023-02-10
feat: custom locale parser

ineshbose:
### 🔗 Linked issue

#1666 

### ❓ Type of change

### 📚 Description

This PR hopes to extend configuration to the module to not assume that locale would be the first part of the file path, but instead, let users add their logic based on their preference/codebase.

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -200,6 +200,16 @@ export interface ModuleOptions {
    */
   defaultLocale?: string
   /**
+   * Custom function to extract the locale for a file.
+   * Note: `_locale` in file takes precedence.
+   *
+   * @default (path, { locales, defaultLocale }) => {
+   *   const parts = path.split('/')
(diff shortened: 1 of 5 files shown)


ineshbose:
Would really appreciate feedback and some advice on how to write test cases please!

farnabaz:
Hey 👋 
Passing function in module options is not a good practice, and it is not a solution for this issue. Module options are going to encode, and encoding functions are not supported.

Instead of introducing a new config in module options, we can extract locale login into a separate transformer. By doing it, users will be able to create a custom transformer to overwrite it and have custom logic for local extraction.

Some notes:
- Locale transformer should remove locale from content `_path`
- The transformer should load before path-meta transformer to make sure that we removed locale from content path before processing `_path`

farnabaz:
Feel free to open PR for it 🙂

PR #1926, opened 2023-02-25
feat: Add ignorePredicate to Config

Neo-Zhixing:
### ❓ Type of change

### 📚 Description

Currently we allow the user to specify a `ignores` list in config which is essentially a list of regular expressions. This is useful but limited. This PR proposes an additional item, `ignorePredicate`, that allows the user to supply a custom callback function to decide if a field should be ignored.

### 📝 Checklist

### DIFF
--- a/docs/content/4.api/3.configuration.md
+++ b/docs/content/4.api/3.configuration.md
@@ -126,6 +126,23 @@ export default defineNuxtConfig({
 })
 ```
 
+## `ignorePredicate`
+
+- Type: `undefined | (string) => bool`{lang=ts}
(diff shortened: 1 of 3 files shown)


farnabaz:
Passing function from `nuxt.config` into Nuxt runtime via runtime config is not a good solution. Runtime configs are plain JSON data which will be store as a json object. AFAIK encoding/decoding functions in runtime config is not yet supported and most possibly never will.

farnabaz:
Closing this PR, Feel free to open an issue to discuss about this feature and figure out workarounds or possible improvements.

PR #1973, opened 2023-03-25
feat(code block): add line number attribute to code block lines

nobkd:
### 🔗 Linked issue

#1972
#1970

### ❓ Type of change

### 📚 Description

This PR adds an attribute `line` for each line of a code block. Inline codes are unaffected.
A line attributes contains the line number for the current line.

Non-highlighted code blocks always just had **one** *line*.
Therefore, a new function was added to split the plain code to individual lines.

### 📝 Checklist


nobkd:
Maybe there's a better way to add those line numbers, so we're also able to display them next to the code as requested in #1970

farnabaz:
> Maybe there's a better way to make those line numbers available, so we're also able to display them next to the code as requested in https://github.com/nuxt/content/discussions/1970

I think it would be better if we add line number as a data attribute instead of class

```
<div data-line="1" ...>...</div>
<div data-line="2" ...>...</div>
```

This way, users can display line number using with CSS styling.

WDYT @nobkd ?

nobkd:
LGTM, then someone should be able to style a line with `[line="X"]` and use the `attr(line)` to append it to `::before`.

nobkd:
@farnabaz
I have updated the code.

I think this should be mentioned in the docs, but I don't quite know where to put it. Do you have any suggestion?
Maybe here? https://content.nuxtjs.org/guide/writing/markdown#code-highlighting

farnabaz (approved):
Thanks

Issue #2007, opened 2023-04-13
`<div>` element is not permitted as content under `<code>`


selfire1:
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.10.0`
- Nuxt Version:     `3.4.0`
- Nitro Version:    `2.3.3`
- Package Manager:  `yarn@1.22.19`
- Builder:          `vite`
- User Config:      `extends`, `modules`, `runtimeConfig`, `app`, `image`, `content`, `nitro`
- Runtime Modules:  `nuxt-icon@0.3.3`, `@nuxt/content@2.5.2`, `@nuxt/image-edge@1.0.0-28020728.5df24eb`, `@nuxtjs/tailwindcss@6.6.5`, `@nuxtjs/html-validator@1.2.4`, `@nuxtjs/color-mode@3.2.0`
- Build Modules:    `-`

### Reproduction

1. Enable `highlight` in the content config

```vue
export default defineNuxtConfig({
  modules: ['@nuxt/content'],
  content: {
    highlight: {
      // https://github.com/shikijs/shiki/blob/main/docs/themes.md
      theme:  'min-light',
    },
  },
});
```

2. Render a fenced code block.

---

* Minimal reproduction on StackBlitz: [StackBlitz](https://stackblitz.com/edit/github-fwhuqq?file=nuxt.config.ts)

### Describe the bug

This is an example of how Content outputs a fenced code block with `highlight` enabled (Example taken from the NuxtContent Website: [queryContent() · Nuxt Content](https://content.nuxtjs.org/api/composables/query-content))

```html
<pre>
  <code>
    <div class="line"><span class="ct-952581">// Create a query looking for anything in content/ directory</span></div>
    <div class="line"><span class="ct-d7952f">const</span><span class="ct-cf3e01"> </span><span
        class="ct-d584cd">contentQuery</span><span class="ct-cf3e01"> </span><span class="ct-d7952f">=</span><span
        class="ct-cf3e01"> </span><span class="ct-b07143">queryContent</span><span class="ct-cf3e01">()</span></div>
    <div class="line"></div>
    <!-- … -->
  </code>
</pre>
```

Using [nuxtjs/html-validator](https://html-validator.nuxtjs.org/), this throws an error for me: `error  <div> element is not permitted as content under <code>   element-permitted-content`.

[...]

PR #2008, opened 2023-04-14
fix(code-block): use span with line-break tags

nobkd:
### 🔗 Linked issue

#2007

### ❓ Type of change

### 📚 Description

Reverts code block lines to be wrapped in `<span>`s.
Iterates through the lines and adds `<br>` line breaks **between** lines (no break at start or end of code block).
Sets the style of `.line` class to `inline-block`, so the `<br>`s use the same line as a line... 🤦

Has no effect on inline code, as it just uses the content of the first line.

### 📝 Checklist


src/runtime/transformers/shiki/highlighter.ts
```diff
@@ -145,17 +145,24 @@ export const useShikiHighlighter = createSingleton((opts?: Exclude<ModuleOptions
...
       props: {
         class: ['line', highlights.includes(lineIndex + 1) ? 'highlight' : ''].join(' ').trim(),
         line: lineIndex + 1
       },
       children: line.map(tokenSpan)
     }))
 
+    for (let i = highlightedLines.length - 1; i > 0; i--) {
+      highlightedLines.splice(i, 0, lineBreak)
+    }
```
farnabaz, line 162: I believe we don't need to add `<br>` tag after each line. These lines will wrap inside `<pre>` tag which respects new lines `\n`

Simply changing `div` to `span` should do the trick

nobkd: But then we have the problem, that copying code by selection in Firefox removes the line breaks, again.
That was the reason to use divs instead of spans.

farnabaz: Adding `<br>` tag will result in duplicate new lines.

Screenshot 2023-04-17 at 14 23 31

I remember that, but I think we should fix that issue on `CopyButton`. WDYT?

nobkd: The copy button does not have that problem, because it gets the raw code.

Only copying by selection in Firefox is problematic, because spans styled as block don't get copied with newlines between the lines

farnabaz: What about appending a line break char (`\n`) to the last token of each line, if there isn't any? Firefox should detect that

nobkd: I'm at it. Thanks

src/runtime/transformers/shiki/highlighter.ts
```diff
@@ -148,12 +148,12 @@ export const useShikiHighlighter = createSingleton((opts?: Exclude<ModuleOptions
...
     return lines.map((line, lineIndex) => ({
       type: 'element',
-      tag: 'div',
+      tag: 'span',
       props: {
         class: ['line', highlights.includes(lineIndex + 1) ? 'highlight' : ''].join(' ').trim(),
         line: lineIndex + 1
       },
-      children: line.map(tokenSpan)
+      children: [...line.map(tokenSpan), { type: 'text', value: '\n' }]
```
nobkd, line 156:
```suggestion
      children: [...line.map(tokenSpan), { type: 'text', value: lineIndex !== lines.length - 1 ? '\n' : '' }]
```
We could also add a check if it is the last line, even though the browser ignores the trailing \n in the last line

src/runtime/transformers/shiki/highlighter.ts
nobkd: Now using line break characters

src/runtime/transformers/shiki/highlighter.ts
farnabaz: I like it, let's prevent adding `\n` to last line.

Also, WDYT about adding `\n` directly into `line[line.length - 1].content`. This will prevent adding one extra tag to HTML

nobkd: LGTM

farnabaz (approved):
LGTM 👍 
Thanks

PR #2022, opened 2023-04-17
feat: more flexible ignores configuration

### LINKED ISSUE #2017: Better `ignores` implementation / documentation
#### Overview

The current ignores implementation has a few issues:

- options are converted to RegExps which only work as folder / file name _prefixes_
- this a) is not made clear in the docs and b) could be avoided
- content watching / refresh does not respect ignored files

#### Implementation

Looking at the ignore code, it means ignore options can only ever be [prefixes](https://github.com/nuxt/content/blob/main/src/module.ts#L646-L661):

```ts
export const contentIgnores: Array<RegExp> = contentConfig.ignores.map((p: any) =>
  typeof p === 'string' ? new RegExp(`^${p}\|:${p}`) : p
```

This makes sense in the terms of the supplied defaults:

```
ignores: ['\\.', '-'],
```

But is limiting in terms of more flexible ignoring (unless you get quite creative with the input strings).

> Additionally: the implementation suggests that RegExps might be able to be passed, but they seem to get nuked by being squeezed through `getRuntimeConfig()`.

#### Docs

The docs actually don't mention the prefixing, they suggest you can ignore _words_:

- https://content.nuxtjs.org/api/configuration#ignores

This should be clarified.

#### File watching

It looks like the `ignores` config doesn't have any effect on the storage instance at the bottom of `src/module.ts`. 

Saving ignored files still triggers `refreshNuxtData()` via a sockets call.

To support the documented ignores format, change the default to:

```ts
const ignores: string[] = [
  '^[.-]|:[.-]'
]
```

And simplify the `contentIgnores` map to:

```ts
const contentIgnores = contentConfig.ignores.map(p => new RegExp(p))
```

Although, if prefixing is a specific design decision, then document it and provide a couple of examples:

```ts
[
  '.+?\\.txt$' // converts to /:.+?\.txt$/, matching text files only
]
```

To prevent triggering a refresh when changing ignored files, [in the module watch](https://github.com/nuxt/content/blob/main/src/module.ts#L701) use the same algorithm to prevent a sockets broadcast:

```ts
      await nitro.storage.watch(async (event: WatchEvent, key: string) => {
        // Ignore events that are not related to content
        if (isIgnored(key)) {
          return
        }
        ...
      })
```

Finally, as the ignore functionality will be used three times at this point, consolidate into a reusable function:

- https://github.com/search?q=repo%3Anuxt%2Fcontent%20contentIgnores&type=code

I've been poring through the `ignores` code in the IDE and debugger, and pretty sure this is the only option.

### Additional context

Trying to optimise the updating code in Nuxt Content Assets:

- https://github.com/davestewart/nuxt-content-assets/issues/30

Pretty sure I know enough about Nuxt Content's source now to do a PR.

davestewart:
### ❓ Type of change

### 📚 Description

This PR enhances Nuxt Content's ignores implementation.

It has three main aims:

- make the options more flexible
- add a single factory function to test ignore patterns
- ensure ignored files don't trigger an application refresh

#### Options

Right now, the `ignores` implementation transforms user-supplied options so that they work with atoms of the `unstorage` path syntax, by converting the string to a RegExp and prefixing it with a `:` delimiter (which equates to `/`).

Thus, `'hidden'` will only ever match `/hidden` and not `/my-hidden-folder`. In order for a user to get around this they would need some additional RegExp acrobatics, i.e. `'.+hidden'`. Ignoring file extensions would require `'.+\\.ext'` rather than the more idiomatic `'\\.ext$'`.

This change gives the user the flexibility to decide how to ignore content, whist respecting the current defaults to ignore `.dot` and `-dash` files.

#### Predicate

The new `makeIgnores()` factory function returns a predicate which eliminates the duplicate code from the rest of the module.

#### Watch

Currently the module's watch functionality doesn't respect ignored files, and triggers an application refresh even when ignored files are saved.

The update uses the predicate function to ignore ignored file updates and not refresh the app.

#### Testing

I'm not 100% sure how to test this functionality using a test framework, but I have updated the `playground/document-driven` to test it out in the browser.

Resolves #2017

### 📝 Checklist


davestewart:
I just generated the document driven playground with the some configured ignore rules, and it doesn't generate, so I guess it works!

image

farnabaz:
Thanks for the PR, 
It looks good to me 👍 

The only concern I have is about the breaking change we are creating with this PR.
Do you see any possible ways of workarounds to prevent the breaking change?
Or maybe detect it and warn users.

davestewart:
That was my only concern too.

Could potentially:

- create a new config option `ignore` for the more flexible functionality
- deprecate old option
- handle both options in the one factory function until next major version

Or:

- restore the `:` delimited functionality
- update documentation examples to include the `'\\.*'` workaround

My hunch is that the existing `ignores` config is too esoteric for anyone to have figured it out otherwise (but that offers no guarantees!).

farnabaz:
> - create an experimental config option ignores for the more flexible functionality
> - handle both options in the one factory function until it is decided to include it

I like this approach more that others.
- We can enable this new feature under the flag.
- In module check if the user registered any `ignores` pattern and warn user about the breaking change in `v2.7`

davestewart:
Aha! It exists!

https://github.com/nuxt/content/blob/main/src/module.ts#L281-L284

OK, I'll get that written now 👍 

Is there any existing format for warnings? Just `console.warn()` or an existing utility function?

farnabaz:
We can change docs a bit to introduce this experimental flag (the ones you've added in the PR).
Then we can use `consola.warn` to briefly note about behaviour change and link to docs.

WDYT?

davestewart:
OK, copying the example here:

- https://github.com/nuxt/nuxt/blob/886350a27c6b733ae07dfd5f6b8457214aa83fe6/packages/nuxi/src/commands/build.ts#L52

```ts
      consola.warn('The `ignores` format is being made more flexible in version 2.7. See the docs for more information: `https://content.nuxtjs.org/api/configuration#ignores')
```

davestewart:
OK! All done.

Sorry it took me so long, but here are the main changes:

#### Format

I made a small modification so the new format now supports `/` directly in the pattern!

This means users won't need to use `:` in place of `/`

#### `makeIgnores()`

This now takes the runtime config _directly_, and `ignores` and `experimental.ignores` patterns are grabbed.

If the user passes experimental options, the new algorithm is used in place of the old one.

#### `defaults`

I've moved the `.dot` and `.dash` defaults to the factory function.

This is because:

- the old vs new prefixes are subtly different
- the normal and experimental options are declared in two different places
- the normal and experimental options are JSDoc'ed in two different places

Additionally, this makes the warning logic less magic; _any_ supplied ignore options (vs > "magic" 2) will be caught.

#### Playground

I moved the ignore examples to `/basic` as it is a basic feature.

I also added `blah` as an ignore option in `/basic` to trigger the warning

##### Docs

The docs now explain how the two versions differ, and how to use the new version.

image

---

I hope that's all clear and good to go!

davestewart:
OK, done!

src/module.ts
```diff
@@ -600,7 +605,8 @@ export default defineNuxtModule<ModuleOptions>({
       integrity: buildIntegrity,
       experimental: {
         stripQueryParameters: options.experimental.stripQueryParameters,
-        clientDB: options.experimental.clientDB && nuxt.options.ssr === false
+        clientDB: options.experimental.clientDB && nuxt.options.ssr === false,
+        ignores: options.experimental.ignores
```
farnabaz, line 609:
```suggestion
        advancedIgnoresPattern: options.experimental.advancedIgnoresPattern
```

src/module.ts
```diff
@@ -632,6 +638,12 @@ export default defineNuxtModule<ModuleOptions>({
       tailwindConfig.content.push(resolve(nuxt.options.buildDir, 'content-cache', 'parsed/**/*.md'))
     })
 
+    // ignore files
+    const isIgnored = makeIgnored(contentContext)
+    if (contentContext.ignores.length && !contentContext.experimental) {
```
farnabaz, line 643:
```suggestion
    if (contentContext.ignores.length && !contentContext.experimental.advancedIgnoresPattern) {
```

farnabaz (approved):
👍

farnabaz:
@davestewart Looks great.

Do you mind, fixing CI?

src/module.ts
davestewart: Whoops!

davestewart:
Re. release, I haven't updated the package version.

davestewart:
BTW, writing some tests for this, this morning.

farnabaz:
I'm planning to release v2.6 today

Also, before that, once we merge this PR you can test it on [edge-channel](https://content.nuxtjs.org/guide/migration/edge-channel)

davestewart:
OK, I've written some test to test various ignores patterns for old and new.

image

The thing I don't get is, how do I run an individual test as each test file exports a closure?

davestewart:
Question, for future reference...

I presume PR commits are squashed, so [as per](https://www.conventionalcommits.org/en/v1.0.0/#do-all-my-contributors-need-to-use-the-conventional-commits-specification) the Conventional Commits docs – do I need to use this style in the feature branch if you'll just merge with `feat: improve ignores` or such like?

It's a bit of a pain choosing which prefix to use with so many small updates...

farnabaz:
Not sure I'm getting your point here. But For me this is an improvement and new feature so I'd use `feat` for it.
As for the branch it does not matter. As you said, PR will be squashed so the only thing matters is PR title

davestewart:
Yeah I meant for those in the branch.

OK, good to know, thanks.

farnabaz (approved):
👍

farnabaz:
Let's wait for CI to finish and you can test in edge channel

davestewart:
FWIW I get a 503 at my end with the edge channel:

```
file:///Volumes/Data/Work/OpenSource/JavaScript/NuxtContentAssets/nuxt-content-assets/demo/.nuxt/dev/index.mjs:56
import { consola } from 'file:///Volumes/Data/Work/OpenSource/JavaScript/NuxtContentAssets/nuxt-content-assets/demo/node_modules/unenv/runtime/npm/consola.mjs';
         ^^^^^^^
SyntaxError: The requested module 'file:///Volumes/Data/Work/OpenSource/JavaScript/NuxtContentAssets/nuxt-content-assets/demo/node_modules/unenv/runtime/npm/consola.mjs' does not provide an export named 'consola'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:123:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:189:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:533:24)
    at async loadESM (node:internal/process/esm_loader:91:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12)
```

Have hacked node modules so I can test the feature.

farnabaz:
Which version of Nuxt are you using?
Did you try removing lock file and reinstalling?

davestewart:
OK, did that and still get it:

```
[12:43:58 PM] ℹ [content-assets] Websocket listening on "ws://localhost:4001/"
[12:44:02 PM] ℹ Vite client warmed up in 2666ms
[nitro 12:44:02 PM] ✔ Nitro built in 1323 ms
[12:44:12 PM] [nuxt] [request error] [unhandled] [500] Named export 'consola' not found. The requested module 'consola' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'consola';
const { consola } = pkg;

  at ModuleJob._instantiate (node:internal/modules/esm/module_job:123:21)  
  at async ModuleJob.run (node:internal/modules/esm/module_job:189:5)  
  at async Promise.all (index 0)  
  at async ESMLoader.import (node:internal/modules/esm/loader:533:24)  
  at async ./demo/.nuxt/dev/index.mjs:4138:24  
  at async ./demo/.nuxt/dev/index.mjs:4217:64  
  at async ./demo/.nuxt/dev/index.mjs:733:22  
  at async Object.handler (./node_modules/h3/dist/index.mjs:1247:19)  
  at async Server.toNodeHandle (./node_modules/h3/dist/index.mjs:1322:7)
[12:44:12 PM] [nuxt] [request error] [unhandled] [500] Named export 'consola' not found. The requested module 'consola' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'consola';
const { consola } = pkg;

  at ModuleJob._instantiate (node:internal/modules/esm/module_job:123:21)  
  at async ModuleJob.run (node:internal/modules/esm/module_job:189:5)  
  at async Promise.all (index 0)  
  at async ESMLoader.import (node:internal/modules/esm/loader:533:24)  
  at async ./demo/.nuxt/dev/index.mjs:4138:24  
  at async ./demo/.nuxt/dev/index.mjs:4217:64  
  at async ./demo/.nuxt/dev/index.mjs:733:22  
  at async Object.handler (./node_modules/h3/dist/index.mjs:1247:19)  
  at async Server.toNodeHandle (./node_modules/h3/dist/index.mjs:1322:7)
```
```
- Operating System: `Darwin`
- Node Version:     `v16.17.1`
- Nuxt Version:     `3.4.1`
- Nitro Version:    `2.3.3`
- Package Manager:  `npm@8.15.0`
- Builder:          `vite`
- User Config:      `app`, `modules`, `content`, `contentAssets`
- Runtime Modules:  `../src/module`, `@nuxt/content-edge`
- Build Modules:    `-`
```

davestewart:
Looks like line 2:

https://github.com/nuxt/content/blame/12d5a21e62b380889bc0767bb08d7c9b18ff26bd/src/runtime/transformers/shiki/highlighter.ts#L2

davestewart:
OK, it looks like I had a reference to `@nuxt/content` in my dev devependencies and that was installing the old version.

Have just nuked everything and rerun and it works!

Bloody computers 😆

davestewart:
Also, pleased to report that the updates mean I can get the next couple of optimisations for Nuxt Content Assets out!

Thanks so much for working through this PR with me. Really appreciated!

PR #2033, opened 2023-04-20
fix(link): Remove hash before checking if ending by '.md'

Skraye:
### ❓ Type of change

Does not know if it's a bug fix or an enhancement.

### 📚 Description

The PR consists of allowing hash in a relative title.

Example of an URL used in our md files : 
`../docs/04.developer-guide/04.inputs.md#set-inputs-with-curl`

Without the hash part, this works fine and the URL is well interpreted, but we notice that having a hash will break the functionality.

In fact, the method `normalizeLink()` works only with relative files that end with `.md`, so we added hash handling in this function, which is just a regex that removes hash, and set it back before returning.

### 📝 Checklist

- [ ] I have updated the documentation accordingly. -> Did not find any documentation about this.

### DIFF
--- a/src/runtime/markdown-parser/handler/link.ts
+++ b/src/runtime/markdown-parser/handler/link.ts
@@ -29,8 +29,10 @@ export default function link (h: H, node: Node) {
 }
 
 function normalizeLink (link: string) {
-  if (link.endsWith('.md') && (isRelative(link) || (!/^https?/.test(link) && !link.startsWith('/')))) {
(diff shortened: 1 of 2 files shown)


farnabaz:
Thanks for the PR @Skraye,
Do you mind adding a test for this?

Skraye:
Hi @farnabaz 
Did update the test !

farnabaz (approved):
👍

PR #2060, opened 2023-05-09
fix(tailwind-integration): allow content as object

ineshbose:
### 🔗 Linked issue

https://github.com/nuxt-modules/tailwindcss/issues/671

### ❓ Type of change

### 📚 Description

Prepares for incoming change to Tailwind module.

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -635,9 +635,7 @@ export default defineNuxtModule<ModuleOptions>({
     // @ts-ignore - Module might not exist
     nuxt.hook('tailwindcss:config', (tailwindConfig) => {
       tailwindConfig.content = tailwindConfig.content ?? []
-      (Array.isArray(tailwindConfig.content) ? tailwindConfig.content : tailwindConfig.content.files).push(
-        resolve(nuxt.options.buildDir, 'content-cache', 'parsed/**/*.md')
-      )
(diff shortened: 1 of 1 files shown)


src/module.ts
```diff
@@ -634,8 +634,8 @@ export default defineNuxtModule<ModuleOptions>({
     // @nuxtjs/tailwindcss support
     // @ts-ignore - Module might not exist
     nuxt.hook('tailwindcss:config', (tailwindConfig) => {
-      tailwindConfig.content = tailwindConfig.content ?? []
-      tailwindConfig.content.push(resolve(nuxt.options.buildDir, 'content-cache', 'parsed/**/*.md'))
+      tailwindConfig.content = tailwindConfig.content ?? [];
+      (Array.isArray(tailwindConfig.content) ? tailwindConfig.content : tailwindConfig.content.files).push(resolve(nuxt.options.buildDir, 'content-cache', 'parsed/**/*.md'))
```
farnabaz, line 638: Are we sure that `tailwindConfig.content.file` always exists if `tailwindConfig.content` is an object? If not, we should handle `undefined` state of `tailwindConfig.content.file`

ineshbose: I'm confident that it will exist (based on type definition and custom config merging strategy in the module), but better to be safe than sorry - made the change!

PR #2066, opened 2023-05-15
chore(deps): Install unist-util-stringify-position

### LINKED ISSUE #2063: Missing `"unist-util-stringify-position"` dependency in `package.json`
### Environment

yarn v1.22

### Reproduction

Use yarn to add another library that has 'unist-util-stringify-position' as a dependency (an earlier major release). TinaCMS uses a library that uses this. Other notes on reproducing [here](https://github.com/tinacms/tinacms/issues/3921).

```
yarn add tinacms 
```

Notice that yarn will hoist the dependency to the top. 

```
cat ./node_modules/unist-util-stringify-position/package.json
```

This will show the older version (2.0.3)
```
{
  "name": "unist-util-stringify-position",
  "version": "2.0.3",
  "description": "unist utility to serialize a node, position, or point as a human readable location",
  "license": "MIT",
...
```

Yarn always hoists when it won't conflict with another version. In this case, if `@nuxt/content` had listed 'unist-util-stringify-position' as a dependency, the older version wouldn't be hoisted.

### Describe the bug

The dependency is hoisted when it shouldn't be

```
file:///Users/jeffsee/code/scratch/tinacms-nuxt-content/.nuxt/dev/index.mjs:29
import { stringifyPosition } from 'file:///Users/jeffsee/code/scratch/tinacms-nuxt-content/node_modules/unist-util-stringify-position/index.js';
         ^^^^^^^^^^^^^^^^^
SyntaxError: Named export 'stringifyPosition' not found. The requested module 'file:///Users/jeffsee/code/scratch/tinacms-nuxt-content/node_modules/unist-util-stringify-position/index.js' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:
```

### Additional context

I think the only thing that's needed here is to add `unist-util-stringify-position` as a dependency, I believe you'll want a version `>3`. It's being used directly here https://github.com/nuxt/content/blob/af7f1fecd59cdec25e2f365dd2afc46b52c56ee7/src/runtime/transformers/csv/from-csv.ts#L8

### Logs

_No response_

adamalfredsson:
### 🔗 Linked issue

fix #2063

### ❓ Type of change

### 📚 Description

`pnpm install unist-util-stringify-position`

Resolves #2063

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "unified": "^10.1.2",
     "unist-builder": "^3.0.1",
     "unist-util-position": "^4.0.4",
+    "unist-util-stringify-position": "^3.0.3",
     "unist-util-visit": "^4.1.2",
     "unstorage": "^1.6.0",
     "ws": "^8.13.0"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -97,6 +97,9 @@ dependencies:
   unist-util-position:
     specifier: ^4.0.4
     version: 4.0.4
+  unist-util-stringify-position:
+    specifier: ^3.0.3
+    version: 3.0.3
   unist-util-visit:
     specifier: ^4.1.2
     version: 4.1.2


farnabaz (approved):
Thanks ❤️

PR #2067, opened 2023-05-16
fix(code-block): meta property

farnabaz:
### ❓ Type of change

### 📝 Checklist


src/runtime/markdown-parser/handler/utils.ts
```diff
@@ -20,11 +20,13 @@ export function parseThematicBlock (lang: string) {
   const language = lang.replace(/[{|[](.+)/, '').match(/^[^ \t]+(?=[ \t]|$)/)
   const highlightTokens = lang.match(/{([^}]+)}/)
   const filenameTokens = lang.match(/\[(.+)\]/)
+  const meta = lang.replace(/^\w+\s*(\[[^\]]+\])?\s*(\{[^}]+\})?\s*/g, '')
```
nobkd, line 23: A problem is, that the meta regex has a fixed order for `lang […] {…} meta`
It is not replaced if we write `lang {…} […] meta`, even though L21-22 allow different ordering.

Also, if we want to use `{…}` or `[…]` in the meta block, but not provide highlights or a filename, it will be matched by L21-22.
Maybe, the user could provide an empty `[]` or `{}` block before the meta to prevent it from reading those parts in the meta.
Something like: `lang [] meta=['not', 'a', 'filename']`

In my opinion, we should also `trim` the `lang` property before checking if it is empty (in L9 / L11) and return an empty string (`''`) for the `meta` if `lang` is empty

src/runtime/markdown-parser/handler/utils.ts
```diff
@@ -20,11 +20,13 @@ export function parseThematicBlock (lang: string) {
   const language = lang.replace(/[{|[](.+)/, '').match(/^[^ \t]+(?=[ \t]|$)/)
   const highlightTokens = lang.match(/{([^}]+)}/)
```
nobkd, line 21:
```suggestion
  const highlightTokens = lang.match(/{([\s\d\-,]+)}/)
```
It might be a good idea to prevent this from matching things that can't be parsed by `parseHighlightedLines` (new L33)
(Maybe also make content optional with `*` instead of `+` to be able to not provide highlights, but still use `{…}` in `meta`)

nobkd (changes requested):

src/runtime/markdown-parser/handler/utils.ts
nobkd, line 23:
```suggestion
  const highlightTokens = lang.match(/{([^}]*)}/)
  const filenameTokens = lang.match(/\[([^\]]*)\]/)
  const meta = lang.replace(/^\w+\s*(\[[^\]]*\]|\{[^}]*\})?\s*(\[[^\]]*\]|\{[^}]*\})?\s*/, '')
```

This suggestion would allow empty `highlight` (`{}`) and `filename` (`[]`) blocks.
Therefore, if a user doesn't want to provide any of those, but use meta / wants to use the characters in the meta part, can do so.

It also makes the order of `highlight` and `filename` blocks arbitrary: `[]{}` and `{}[]` are both possible when replacing them for the meta.

nobkd (changes requested):

src/runtime/markdown-parser/handler/utils.ts
nobkd, line 23: Previous comment:

A problem is, that the `meta` regex has a fixed order for `lang […] {…} meta`
It is not replaced if we write `lang {…} […] meta`, even though **L21-22** allow different ordering.

Also, if we want to use `{…}` or `[…]` in the `meta` block, but not provide `highlights` or a `filename`, it will be matched (by **L21-22**).
Maybe, the user could provide an empty `[]` or `{}` block before the `meta` to prevent it from reading those parts in the meta.
Something like: `lang [] meta=['not', 'a', 'filename']` would be possible then

In my opinion, we should also `trim` the `lang` property before checking if it is empty (adding it in **L11**) and return `undefined` for the `meta` if `lang` is empty (adding in **L17**)

nobkd (approved):
Looks good to me.

But I don't know if it's good to return empty strings on empty `filename`, or if we should return `undefined` instead... 🤔

What do you think?

src/runtime/markdown-parser/handler/utils.ts
```diff
@@ -9,22 +9,25 @@ export function parseThematicBlock (lang: string) {
...
-  const filenameTokens = lang.match(/\[(.+)\]/)
+  const highlightTokens = lang.match(/{([^}]*)}/)
+  const filenameTokens = lang.match(/\[([^\]]*)\]/)
+  const meta = lang.replace(/^\w*\s*(\[[^\]]*\]|\{[^}]*\})?\s*(\[[^\]]*\]|\{[^}]*\})?\s*/, '')
 
   return {
     language: language ? language[0] : undefined,
     highlights: parseHighlightedLines(highlightTokens && highlightTokens[1]),
-    filename: Array.isArray(filenameTokens) ? filenameTokens[1] : undefined
+    filename: Array.isArray(filenameTokens) ? filenameTokens[1] : undefined,
```
nobkd, line 29:
```suggestion
    filename: Array.isArray(filenameTokens) && filenameTokens[1] ? filenameTokens[1] : undefined,
```

nobkd, line 29:
```suggestion
    filename: Array.isArray(filenameTokens) && filenameTokens[1] ? filenameTokens[1] : undefined,
```

test/features/parser-markdown.ts
```diff
@@ -58,12 +58,62 @@ export const testMarkdownParser = () => {
...
+      })
+
+      expect(parsed).toHaveProperty('body')
+      expect(parsed.body).toHaveProperty('children[0].tag', 'code')
+      expect(parsed.body).toHaveProperty('children[0].props')
+      const props = parsed.body.children[0].props
+      expect(props).toHaveProperty('meta')
+      expect(props.meta).toBe('other code block info')
+      expect(props.language).toBe(undefined)
+      expect(props.filename).toBe('')
```
nobkd, line 113:
```suggestion
      expect(props.filename).toBe(undefined)
```

nobkd (commented):
This would be my change, ***if*** we don't want to return empty strings.

PR #2075, opened 2023-05-21
feat: add i18n support

NftTopBest:
### 🔗 Linked issue

i18n feature issue item from docus: https://github.com/nuxt-themes/docus/issues/770

### ❓ Type of change

### 📚 Description

The PR adds i18n features.
The idea comes from here:  https://github.com/nuxt-themes/docus/issues/770
After some more deep digs, I find out the solution is to make a PR here. And nothing to do with the docus project.

Screenshot 2023-05-21 at 21 39 49

- [x] Add `playground/i18n` example
- [x] Add `dev:i18n` command in `package.json`

### 📝 Checklist


Tahul:
@farnabaz @Atinux ;

This looks reasonable, I briefly checked the code and it seem to only be built for Document Driven mode.

As I think this behavior can easily be achieved via regular prefixed content queries, what do you think in including this in Document Driven mode?

We could also move that code somewhere in Docus I guess.

Atinux:
I don't think `useI18n()` is the correct name, I would at least name it `useContentI18n()` to avoid conflict with other libraries.

Before changing the document driven mode, I believe we need first to have an upgrade of it to avoid using a middleware but a plugin to inject the page to vue-router itself, this way we can leverage payload extraction as well as better i18n support.

NftTopBest:
> We could also move that code somewhere in Docus I guess.
This is why we can not make it in Docus.

1. At the beginning, I try to make it in Docus, then find out that I have to update the code here
2. And finally all the code setup in Content is enough, no more code for Docus then. 
3. And only changing the code in Docus can not have the ability to update the navigation update to a new lang, check the code here: https://github.com/nuxt/content/pull/2075/files#diff-655e26ed4032bc6928b3b3838a699902e421b1838fadf34ed6d78decdfdfa69dR52-R64

The target features:

1. user can have a lang switcher to switch to multiple langs with the top and sidebar nav change correspond
4. user just simply setup the locales and defaultLocale in the `nuxt.config.ts` to turn on the feature in Docus or any other nuxt app that use the Content, there is no need to install the nuxt-i18n module
5. nuxt-i18n module maybe not suit for this as it seems auto-generate new lang route for every lang, and in the content case

productdevbook:
This issue is extremely important to me, I hope it will be resolved for a moment. Thank you @NftTopBest pr.

truesteps:
Any progress?

NftTopBest:
> 

@truesteps 

Check out the demo  here: https://nuxt-content-i18n.vercel.app/

For now, I am not sure what I should to further.

truesteps:
@NftTopBest I love the implementation, I meant any progress on the maintainers side :) thanks for the quick work, it looks amazing

XenBG:
I'd like to use that soon.

Barbapapazes:
Whooo, awesome. Thanks for this! Can't wait to try it in my portfolio because it's currently written in French but many wants to read it in English.

NftTopBest:
Hi @farnabaz , is there anything I need to modify for the code?

Barbapapazes:
> Before changing the document driven mode, I believe we need first to have an upgrade of it to avoid using a middleware but a plugin to inject the page to vue-router itself, this way we can leverage payload extraction as well as better i18n support.

Hello @Atinux, could you explain more "inject the page to vue-router"? Thanks

Barbapapazes:
> Hi @farnabaz , is there anything I need to modify for the code?

Hello, I think that the support for i18n could also be add for non DocumentDriven mode. Using ContentDoc, ContentFetch and ContentNavigation.

XeniaLu:
Hello, maintainers. May I inquire if there have been any recent updates regarding the timeline for merging this feature?

DraftProducts (changes requested):
Another change is needed to support i18n : the [parseMarkdown()](https://github.com/nuxt/content/blob/a93293f3aee2033e3135b4f3cad6b0caa9a266ec/src/runtime/transformers/markdown.ts#L19-L33) function (L19-L33) should be called with some i18n options for the `remark-rehype` plugin :
```diff  
const parsed = await parseMarkdown(content as string, {
      highlight: options.highlight,
      remark: {
        plugins: config.remarkPlugins
      },
      rehype: {
        options: {
+         footnoteLabel: 'Footprints'
          handlers: {
            link: link as any
          }
        },
        plugins: config.rehypePlugins
      },
      toc: config.toc
    }
})
```
ℹ️ The footnote option is documented here : https://github.com/remarkjs/remark-rehype#fields

The current way to configure without i18n is through the `nuxt.config.ts` file with redefining the plugin to force options :
```
content: {
  markdown: {
    remarkPlugins: [
      ['remark-rehype', {
        footnoteLabel: 'Footprints'
      }]
    ]
  }
},
  ```

**Comment by narr07** [comment]
Any progress on this?

**Comment by mthmcalixto** [comment]
Any progress on this?

**Comment by iBobik** [comment]
My current workaround:
```
const { data: page } = await useAsyncData('page:' + path, () => {
  return queryCollection('pages').path(`/pages/${(slug as Array<string>).join('/')}.${locale.value}`).first()
}, { watch: [locale] })
if (!page.value) throw createError({ statusCode: 404, message: 'Page not found' })
```

I choose the lang suffix because this helps us to keep consistency between languages - files are next to each other so easy to navigate and keep in mind:
Snímek obrazovky 2025-05-21 v 14 36 35

**Comment by narr07** [comment]
> My current workaround:
> 
> ```
> const { data: page } = await useAsyncData('page:' + path, () => {
>   return queryCollection('pages').path(`/pages/${(slug as Array<string>).join('/')}.${locale.value}`).first()
> }, { watch: [locale] })
> if (!page.value) throw createError({ statusCode: 404, message: 'Page not found' })
> ```
> 
> I choose the lang suffix because this helps us to keep consistency between languages - files are next to each other so easy to navigate and keep in mind: Snímek obrazovky 2025-05-21 v 14 36 35

Can you provide an example of the folder structure for the pages? Then can you give a full code example for a page?

**Comment by iBobik** [comment]
> Can you provide an example of the folder structure for the pages? Then can you give a full code example for a page?

Folder structure is on the screenshot. Code later

**Comment by iBobik** [comment]
> Can you provide an example of the folder structure for the pages? Then can you give a full code example for a page?

```
<script lang="ts" setup>
const { path, params: { slug } } = useRoute()
const { locale } = useI18n()

const { data: page } = await useAsyncData('page:' + path, () => {
  return queryCollection('pages').path(`/pages/${(slug as Array<string>).join('/')}.${locale.value}`).first()
}, { watch: [locale] })
if (!page.value) throw createError({ statusCode: 404, message: 'Page not found' })
</script>

<template>
  <Head>
    <Title>{{ page!.title }}</Title>
  </Head>

  <NavBar />

  <DonateContainer class="py-10" prose>
    <h1>{{ page!.title }}</h1>
    <ContentRenderer :value="page!" />
  </DonateContainer>

  <Footer />
</template>

```

Deeyuude:
Hello and sorry for the question ! Is this going to use @nuxtjs/i18n module or something Else ?

farnabaz:
Nuxt Content can be integrated with the i18n module. Check out the [i18n example](https://github.com/nuxt/content/tree/main/examples/i18n) for more details.  

A brief guide on integrating with the i18n module is also available in the [official documentation](https://content.nuxt.com/docs/integrations/i18n).

PR #2114, opened 2023-06-20
fix(markdown): allow plugin with array type option

maximepvrt:
### ❓ Type of change

### 📚 Description

The current implementation of the markdown plugins 'rehype' and 'remark' transforme the option array into object. When providing an array, such as `[{selector: 'ol', wrapper: 'p' }, {selector: 'ul', wrapper: 'p' }]`, it is transformed into an object with numeric keys instead of maintaining the original array structure.

```
content: {
        markdown: {
            rehypePlugins: {
                'rehype-wrap-all': [{selector: 'ol', wrapper: 'p' }, {selector: 'ul', wrapper: 'p' }]
            }
        }
    },
```

### 📝 Checklist


src/runtime/markdown-parser/content.ts
```diff
@@ -12,8 +12,7 @@ import { nodeTextContent } from './utils/node'
 const usePlugins = (plugins: Record<string, false | MarkdownPlugin>, stream: Processor) => {
   for (const plugin of Object.values(plugins)) {
     if (plugin) {
-      const { instance, ...options } = plugin
-      stream.use(instance, options)
+      stream.use(plugin.instance, plugin.options)
```
farnabaz, line 15: This will introduce a breaking change for standalone usage.

I think it is better to create a special case for `options` field and add a warning. We may remove old syntax in next releases.

```suggestion
      const { instance, options, ...deprecatedOptions } = plugin
      if (Object.keys(deprecatedOptions).length) {
        console.warn('[Markdown] Deprecated syntax. Please use `options` key in order to pass option to remark/rehype plugins.')
      }
      stream.use(instance, options ? options : deprecatedOptions)
```

farnabaz (changes requested):
Thanks for the PR @maximepvrt 
Do you mind adding a simple test for it?

src/runtime/markdown-parser/content.ts
maximepvrt: @farnabaz sorry, I accepted your suggestion without renaming the commit 🙈

maximepvrt:
@farnabaz What test would you like me to add? Adding a plugin with a configuration of type 'array' in this file: https://github.com/nuxt/content/blob/main/test/fixtures/basic/nuxt.config.ts ? (for example rehype-wrap-all ?)

farnabaz:
Yes using`rehype-wrap-all` would be fine. 
If adding `rehype-wrap-all` in nuxt.config breaks other tests, feel free to duplicate `fixtures/basic/server/api/parse.ts` and add plugin directly there, and use new API for the test

maximepvrt:
@farnabaz updated 🚀

maximepvrt:
@farnabaz Ready to merge. Looks good to me. 🚀

farnabaz (approved):
Thanks ❤️

Issue #2100, opened 2023-06-13
[Performance] Improve Nuxt content performances


Hebilicious:
### Is your feature request related to a problem? Please describe

Whenever I see a website built with Nuxt content, I notice that the page transitions are noticeably slow. 
You can use [Nuxt UI Labs](https://ui.nuxtlabs.com/getting-started) or the Nuxt docs as an example. 
After a refresh, on each page navigation, there's a server delay that can take up to 300ms for me before the page is being displayed, giving the feeling of a laggy application. 

Some resources can even take longer (when there's shiki loaded on the page).

Example :
<img width="484" alt="NuxtContent lag" src="https://github.com/nuxt/content/assets/13395944/82e27a6d-0f98-41a1-8d74-988e534b4f87">
 
### Describe the solution you'd like
As these website are mostly static, they should be very performant and there shouldn't be a visible delay between page transitions.

### Describe alternatives you've considered

I don't have an alternative, as Nuxt Content has many great features.
For markdown only, I use Vitepress and I made a POC version for nuxt :  https://github.com/Hebilicious/nuxtpress

### Additional context

Note that this apply on the latest versions of all browsers in incognito mode. This could be also due to the fact that there's no Caching/CDN for the JSON.

ryoheiw:
@Hebilicious Thank you for writing this issue. i’ve been dealing with this problem for a while now without understanding why. i use both markdown and json and my static site is much slower than I would like.  I guess it was the lack of caching on JSON. Thanks.

Hebilicious:
> @Hebilicious Thank you for writing this issue. i’ve been dealing with this problem for a while now without understanding why. i use both markdown and json and my static site is much slower than I would like. I guess it was the lack of caching on JSON. Thanks.

Could you share what you've done to cache it ? I think it could be useful to others.

ryoheiw:
@Hebilicious I do not have a solution to this problem. The comment above was thanking you for letting me understand the reason for its slowness. I will add something here if I find a solution though.

danielroe:
I totally agree.

I believe that we need to prefetch the components used in document-driven mode. By default Nuxt will prefetch the payload required - which will include an AST of the file. We also expose a `prefetchComponents` utility that Nuxt Content could use to prefetch the components _used_ in this AST.

Let me know if someone fancies tackling this - I'd be happy to help.

Hebilicious:
[...]

PR #2118, opened 2023-06-24
perf: keep document-drive state in `shallowRef` and prefetch pages/components

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/content/issues/2100

### ❓ Type of change

### 📚 Description

This PR implements two changes to speed up the experience on document-driven sites. **It should be considered a proof-of-concept and likely warrants further testing**.

1. First, we take advantage of the new rich JSON payloads to avoid having fully reactive page data. For deeply nested arrays of objects, this is likely to make a huge performance impact.

2. Second, we preload content queries _and_ the components used on those pages (if the `body` attribute is present).

In order to enable testing this out on a large site/codebase, I've deployed a patched version of nuxt content:

* without changes: https://nuxt.com/docs
* with changes: https://nuxt-website-idcnshvbq-nuxt-js.vercel.app/docs

(Bear in mind that to test out the document driven changes you will need to stay within the `/docs` subpath as other pages do not use this strategy and will not be affected by these improvements.)

### 📝 Checklist


farnabaz (approved):
Thanks

rikzwarthoff:
@danielroe @farnabaz @Atinux, Hi guys looks like this is causing some problems. It's also prefetching urls that are relative but external to the application. https://github.com/nuxt/content/issues/2336

Issue #2111, opened 2023-06-17
internal component ProseCode class attribute seems to be not working


narasimhajupally:
### Environment

- Operating System: Linux
- Node Version:     v18.16.0
- Nuxt Version:     3.5.3
- Nitro Version:    2.4.1
- Package Manager:  npm@9.6.6
- Builder:          vite
- User Config:      modules, colorMode, css, content, devtools
- Runtime Modules:  @nuxtjs/tailwindcss@6.7.2, @nuxtjs/color-mode@3.2.0, @nuxt/content@2.7.0, nuxt-icon@0.4.1, @nuxt/image@1.0.0-rc.1
- Build Modules:    -

### Reproduction

We can the following warning
```
Vue warn]: Extraneous non-props attributes (class) were passed to component but could not be automatically inherited because component renders fragment or text root nodes. 
  at <ProseCode class="language-javascript" code="var num1, num2, sum;\nnum1 = prompt(\"Enter first number\");\nnum2 = prompt(\"Enter second number\");\nsum = parseInt(num1) + parseInt(num2); // \"+\" means \"add\"\nalert(\"Sum = \" + sum); // \"+\" means combine into a string\n" language="javascript"  ... > 
  at <AsyncComponentWrapper class="language-javascript" code="var num1, num2, sum;\nnum1 = prompt(\"Enter first number\");\nnum2 = prompt(\"Enter second number\");\nsum = parseInt(num1) + parseInt(num2); // \"+\" means \"add\"\nalert(\"Sum = \" + sum); // \"+\" means combine into a string\n" language="javascript"  ... > 
  at <ContentRendererMarkdown value=
```
clone this repo `https://github.com/narasimhajupally/tailwind-nuxtjs-starter-blog` and go to route `/blog/code-sample`
you will see the warning [Vue warn] in the developer console.

### Describe the bug

classes like `language-javascript` and `language-python` that are needed for syntax highlighting are not coming into effect.`

nobkd:
I also experience this issue on a fresh project without any language set for the code block.

helltraitor:
Inline code highlight maybe enabled with ``` `<code>`{lang=<lang>} ```
```
`const sum = parseInt('1')`{lang=js}
```

```
Vue warn]: Extraneous non-props attributes (class) were passed to component but could not be automatically inherited because component renders fragment or text root nodes. 
[...]

PR #2121, opened 2023-06-26
fix: ProseCode class attribute warning

bobby169:
https://github.com/nuxt/content/issues/2111

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/runtime/components/Prose/ProseCode.vue
+++ b/src/runtime/components/Prose/ProseCode.vue
@@ -22,17 +22,13 @@ export default defineComponent({
     meta: {
       type: String,
       default: null
-    },
-    class: {
-      type: String,
-      default: "",
     }
   }
 })
 </script>
 
 <template>
-  <div :class="class">
+  <div>
     <slot />
   </div>
 </template>
--- a/src/runtime/components/Prose/ProseCode.vue
+++ b/src/runtime/components/Prose/ProseCode.vue
@@ -22,13 +22,19 @@ export default defineComponent({
     meta: {
       type: String,
       default: null
+    },
+    class: {
+      type: String,
+      default: "",
     }
   }
 })
 </script>
 
 <template>
-  <slot />
+  <div :class="class">
+    <slot />
+  </div>
 </template>
 
 <style>


narasimhajupally:
we can just wrap the slot in <div> making it root node, doing so the class prop will be automatically inherited.

src/runtime/components/Prose/ProseCode.vue
```diff
@@ -22,13 +22,19 @@ export default defineComponent({
     meta: {
       type: String,
       default: null
+    },
+    class: {
+      type: String,
+      default: "",
```
nobkd, line 28:
```suggestion
```

src/runtime/components/Prose/ProseCode.vue
```diff
@@ -22,13 +22,19 @@ export default defineComponent({
...
     }
   }
 })
 </script>
 
 <template>
-  <slot />
+  <div :class="class">
```
nobkd, line 35:
```suggestion
  <div>
```

bobby169:
> we can just wrap the slot in `<div>` making it root node, doing so the class prop will be automatically inherited.

I try to remove `:class="class"` from root div element, the  class will not be automatically inherited.

image


Add `:class="class"` the result is:

image

With the `class` prop We can define more styling code.

farnabaz:
Thanks for the PR.
Instead of wrapping code, a better solution would be moving class attribute from component itself into inner `code` tag (` pre > code`)

from:
https://github.com/nuxt/content/blob/d7bccdc5ba4c5348d30ed49ba7ea6cf43cc19462/src/runtime/markdown-parser/handler/code.ts#L27-L29

to:
```ts
[h(node, 'pre', {}, [h(node, 'code', { __ignoreMap: '', className: [`language-${language}`] }, [u('text', code)])])]
```

wJoenn:
@farnabaz I just tried your solution locally and it doesn't solve Vue's warning.
Bobby's fix by wrapping the slot inside a div does on the other hand.

Wouldn't it be just fine to merge this as is ?

It could even serve a second purpose by adding a span with the filename to be able to display the filename in the corner of the code block
```
<template>
  <div class="codeBlock">
    <span class="filename">{{ filename }}</span>
    <slot />
  </div>
</template>

<style>
.codeBlock {
  position: relative;
}

.codeBlock .filename {
  font-size: 0.8em;
  position: absolute;
  right: 20px;
  top: 10px;
}

.codeBlock pre code .line {
  display: block;
  min-height: 1rem;
}
</style>
```

image

farnabaz:
Having filename indeed is a good thing to have, but also some user may not want to have a wrapper tag. 
As I checked a couple of websites, they add the class into `pre` tag. Check https://shiki.matsu.io/

Also, I've checked it locally, adding class to `pre` tag resolves the warning 🙂

1714080902120:
I got the same problem, i updated my `@nuxt/content` version to the latest just now, but it still not work.
image

farnabaz:
I'm closing this in favor of #2187.
This issue is fixed in nuxt-mdc package

PR #2142, opened 2023-07-05
feat: allow to disable content head

### LINKED ISSUE #2141: Allow to disable automatic usage of `useContentHead`

Currently, it's not clear if `nuxt/content` must provide SEO out-of-the box. Current status is in the middle of no SEO and SEO ready (and that's strange but clearly understandable when we look back).

But this behavior can interfere with others modules like nuxt-seo-kit (and submodules) from @harlan-zw.

I think that having this little default is great but for advanced users, allowing to disable completely usage of `useContentHead` could allow to setup custom SEO in the application.

Barbapapazes:
### 🔗 Linked issue

fix #2141

### ❓ Type of change

### 📚 Description

I added a flag to disable usage of `useContentHead` in the module for advanced users.

### 📝 Checklist


src/module.ts
```diff
@@ -203,6 +203,12 @@ export interface ModuleOptions {
    * @default undefined
    */
   defaultLocale?: string
+  /**
+   * Ignore automatic usage of `useContentHead`
```
Barbapapazes, line 207:
```suggestion
   * Enable automatic usage of `useContentHead`
```

src/runtime/components/ContentDoc.vue
```diff
@@ -94,12 +97,12 @@ export default defineComponent({
         // Default slot
         default: slots?.default
           ? ({ data, refresh, isPartial }: any) => {
-              if (head) { useContentHead(data) }
+              if (head && enableContentHead) { useContentHead(data) }
```
farnabaz, line 100: I think it is a better approach if we use `enableContentHead` as default value `head` props and let users overwrite it in special cases if they want to.

src/module.ts
```diff
@@ -746,6 +754,8 @@ interface ModulePublicRuntimeConfig {
 
   navigation: ModuleOptions['navigation']
 
+  enableContentHead: ModuleOptions['enableContentHead']
```
farnabaz, line 757: Maybe `contentHeadMeta` or `contentMetaTags`, instead of `enableContentHead`. WDYT?

src/runtime/components/ContentDoc.vue
Barbapapazes: I don't understand this comment. Could you explain `we use enableContentHead as default value head props`.

src/module.ts
farnabaz: Yes as a boolean, or maybe simpler `contentHead`

src/runtime/components/ContentDoc.vue
farnabaz: This component accepts a `head` props with `true` as default, we can set the default to `undefined` then assing `enableContentHead` to it if it's undefined

maybe this snippet describes it better
```ts
const shouldInjectContentHead = props.head === undefined ? enableContentHead : props.head

if (shouldInjectContentHead) { useContentHead(data) }

```

Barbapapazes: Okkk, I understand. Will update the PR.

src/module.ts
farnabaz: Like `documentDriven` option, users can set it to false

docs/content/4.api/3.configuration.md
```diff
@@ -366,6 +366,13 @@ Default locale for top level contents. Module will use first locale code from `l
 Note that in case of defining multiple locales, Module will filter content with `defaultLocale`. If you want to fetch contents of another locale, you need to use `where` option.
 ::
 
+## `contentHead`
+
+- Type: `Boolean`{lang=ts}
```
nobkd, line 371:
```suggestion
- Type: `boolean`{lang=ts}
```

nobkd (changes requested):

farnabaz (approved):

farnabaz:
Thanks ❤️

PR #2146, opened 2023-07-06
feat: add search

### LINKED ISSUE #553: search by index using fuse.js

Currently, searching through content is possible using `where` query.  it requires loading entire `db.json` on client-side which is big and also is not smart enough when user input is not exactly matching. 

 We can utilize [fuse.js](https://fusejs.io/) to create small index chunks and do fuzzy search.

`nuxt.config`

```js
export default {
  content: {
    indexes: {
       modules: [ 'name', 'description', 'author.name', 'tags' ] 
   }
 }
}
```

```js
await $content.search('modules', 'http')
```

Although it is possible using services like Algolia (#523) and in fact it is more efficient overall, it requires depending on an external service and custom setup. Even if docs supports it, the main content module goal was to remove dependency on API endpoints so this solution would be still useful.

### Additional context

Not fully related, but we can also create an auto index using fuzzy.js based on crawler URLs to detect 404 typos like when one goes to `/api/options` and if we moved to `/options`, we can show a friendly message and link "Did you meant /options?"`

### LINKED ISSUE #1507: Searching for a text in markdown
Hi, 
I am not seeing this in the docs, or maybe its not apparent to me.

Is it possible to search for a text anywhere if I had markdown files?

Say I have a bunch of markdown `.md` files in my `content` dir, and I would like to present a separate page to search anywhere in these md files? 

In my case: 
- user would search by search term
- that searches anywhere in `.md` files in `content` directory
- I preset user the list of page titles and a little excerpt from the page where match was found

Questions:
- if this is presently achievable, and if yes, how?
- inf not, if its possible to put it on a feature request list?

### LINKED ISSUE #1758: .search() for query (nuxt-content v2)

In nuxt-content v1 there was the possibility to search, for example, articles using `.search()`, there is no such thing in v2

```javascript
const search = "javascript"
const articles= await queryContent('/articles/').search(search).only(['title', 'description',' date').find()
```
or if the developer doesn't want to search by the entire contents of the file 
```javascript
const articles= await queryContent('/articles/').search(search, ['title','description']).only(['title', 'description',' date').find()
```

I am currently using something like this (there is no mention of regex in the documentation, but it works). The problem is that I would like to search not only by title but also, for example, by description or even by the contents of the .md file itself.
```javascript
.where({title: { $regex: `/${search}/ig` }})
```

Barbapapazes:
### 🔗 Linked issue

fix #1758
fix #1507
fix #553

### ❓ Type of change

### 📚 Description

This PR add support full-text search.

- Support MiniSearch with or without index
- Return json content to use with any full-text-search

Appreciate feedback 🙏 to improve usage. Once validate, I will write the documentation.

An example of usage : https://github.com/Barbapapazes/esteban-soubiran.site/pull/477

_Heavily inspired by the work of @Tahul in https://github.com/nuxt-themes/docus/pull/930 and the local search of VitePress_

### 📝 Checklist


src/module.ts
```diff
@@ -190,6 +190,44 @@ export interface ModuleOptions {
...
+    /**
+     * Used to determine how to search content.
+     */
+    mode: 'meta' | 'full-text'
+
+    filter: {
+      /**
+       * Extensions to keep for text extraction.
+       *
+       * @default []
```
Barbapapazes, line 208: Default is [md]

Barbapapazes:
I'm looking for a better way to return content in order to allow access to sub title.

Barbapapazes:
> I'm looking for a better way to return content in order to allow access to sub title.

I think that I will pivot to Minisearch https://lucaongaro.eu/blog/2019/01/30/minisearch-client-side-fulltext-search-engine.html

src/runtime/server/api/search.ts
```diff
@@ -0,0 +1,148 @@
...
+  if (node.type === 'text') {
+    text += node.value
+  }
+
+  // Do not explore children
+  if (ignoredTags.includes(node.tag ?? '')) {
+    return ''
+  }
+
+  if (node.tag === 'style') {
```
Barbapapazes, line 136: Remive this since it's in the default.

src/module.ts
```diff
@@ -190,6 +190,31 @@ export interface ModuleOptions {
...
+    /**
+     * List of tags where text must not be extracted. Only works with `full-text` mode.
+    *
+    * By default, will extract text from each tag.
+    *
+    * @default ['style', 'code']
+    */
+    ignoredTags?: Array<string>
+    /**
+     * API return indexed contents to improve client-side load time.
```
Barbapapazes, line 212: Add the fact that using this, dev will need to use MiniSearch (or use search). Disabling this, and dev can use another system like fuse.js

package.json
```diff
@@ -88,6 +89,8 @@
     "@nuxtjs/eslint-config-typescript": "latest",
     "@types/mdurl": "^1.0.2",
     "@types/ws": "^8.5.5",
+    "@vueuse/core": "^10.2.1",
+    "@vueuse/nuxt": "^10.2.1",
```
farnabaz, line 93: I think we should move this to `dependencies` since it is registered via `module.ts`

src/module.ts
```diff
@@ -347,6 +400,43 @@ export default defineNuxtModule<ModuleOptions>({
...
       )
 
+      if (options.search) {
+        if (options.search.indexedSearch) {
+          // Will return a string
+          nitroConfig.handlers.push({
+            method: 'get',
+            route: nuxt.options.dev
+              ? `${options.api.baseURL}/indexed-search.txt`
+              : `${options.api.baseURL}/indexed-search.${buildIntegrity}.txt`,
```
farnabaz, line 410: Is there any specific reason for using `.txt` and not `.json`? 

Also, what do you think about merging `indexed-search.txt` and `search.json`. The result could vary based on `indexedSearch`

src/runtime/composables/search.ts
```diff
@@ -0,0 +1,100 @@
...
+  const { results } = useIndexedMiniSearch(search, data as unknown as string, options)
+
+  return results
+}
+
+export const defineMiniSearchOptions = <DataItem>(options: MiniSearchOptions<DataItem>) => {
+  return ref(options)
+}
+
+export const useSearch = async <DataItem>(search: MaybeRefOrGetter<string>, options: MiniSearchOptions<DataItem>) => {
```
farnabaz, line 23: Would be great if we merge `useIndexedSearch` and `useSearch` into a single composable. Having two composables make no sense because the API works based on `indexedSearch` boolean

src/runtime/server/api/search.ts
```diff
@@ -0,0 +1,20 @@
+import { defineEventHandler } from 'h3'
+import { splitPageIntoSections } from '../search'
+import { useRuntimeConfig } from '#imports'
+import { serverQueryContent } from '#content/server'
+
+export default defineEventHandler(async (event) => {
+  const runtimeConfig = useRuntimeConfig()
+  const { ignoredTags } = runtimeConfig.public.content.search
+
+  const files = await serverQueryContent(event).find()
```
farnabaz, line 10: It is safe if we filter `partial` and `drafted` contents from search by default.
As an advanced option, we can also provide config in module option to control behavior of this search by providing query

farnabaz (changes requested):
Thanks for the PR @Barbapapazes 
I like the idea of pre-rendering the search index to reduce the payload 👍 

Beside my comments, we also need to consider having a solution for `ssr: false`.

package.json
Barbapapazes: Yes, it could be better.

src/module.ts
Barbapapazes: Yes! But maybe I could use some headers to force Nitro to return a json without not in a `html`.

I do this to avoid that Nitro render the string in a `.html` file.

src/runtime/server/api/search.ts
Barbapapazes: Yes, I will add `partial` and `draft` by default and will see later if it's useful to use an option for this behavior.

Barbapapazes:
> Beside my comments, we also need to consider having a solution for `ssr: false`.

In that case, do we have SSG?

package.json
Barbapapazes: Done

src/module.ts
Barbapapazes: Done. I remove txt and update `responseType` when fetching an indexed search.

src/runtime/composables/search.ts
Barbapapazes: I use runtimeConfig!

src/runtime/server/api/search.ts
Barbapapazes: I added some options.

playground/search/components/search.vue
```diff
@@ -0,0 +1,28 @@
+<script lang="ts" setup>
+const search = ref('')
+
+// Usage for indexed search
```
Barbapapazes, line 4:
```suggestion
// Usage for non indexed search
```

playground/search/nuxt.config.ts
```diff
@@ -0,0 +1,17 @@
+export default defineNuxtConfig({
+  ssr: false,
+
+  extends: ['../shared'],
+
+  content: {
+    search: {
+      mode: 'full-text',
+      ignorePartials: true,
+      ignoreEmpty: true,
```
Barbapapazes, line 10:
```suggestion
      ignoreEmpty: true
```

src/runtime/server/api/search.ts
```diff
@@ -0,0 +1,24 @@
+import { defineEventHandler } from 'h3'
+import { splitPageIntoSections } from '../search'
+import { useRuntimeConfig } from '#imports'
+import { serverQueryContent } from '#content/server'
+
+export default defineEventHandler(async (event) => {
+  const runtimeConfig = useRuntimeConfig()
+  const { ignoredTags, ignoreDrafts, ignorePartials, ignoreEmpty, options } = runtimeConfig.public.content.search
```
Barbapapazes, line 8:
```suggestion
  const { ignoredTags, ignoreDrafts, ignorePartials, ignoreEmpty } = runtimeConfig.public.content.search
```

Barbapapazes:
> Thanks for the PR @Barbapapazes I like the idea of pre-rendering the search index to reduce the payload 👍
> 
> Beside my comments, we also need to consider having a solution for `ssr: false`.

And do you think I can start writing docs?

Atinux:
Would you be up to fix the conflicts and write the docs for it @Barbapapazes ?

src/runtime/server/api/indexed-search.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  const files = await serverQueryContent(event).find()
+
+  // Only works for MD
+  const sections = (await Promise.all(
+    files
+      .filter(file => file._extension === 'md' &&
+      (!ignoreDrafts ? true : !file?._draft) &&
+      (!ignoreEmpty ? true : !file?._empty) &&
+      (!ignorePartials ? true : !file?._partial)
```
Atinux, line 19: Could we also support some kind of `ignoreNavigationFalse`?

When a page has:
```md
---
navigation: false
---
```

src/module.ts
```diff
@@ -429,6 +543,50 @@ export default defineNuxtModule<ModuleOptions>({
...
+              titles: 1
+            }
+          }
+        }
+      }
+
+      options.search = {
+        ...defaultSearchOptions,
+        ...options.search
+      }
```
farnabaz, line 572: Using `defu`  users will be able to just change a single field in a nested object without need to provide all fields
```suggestion
      options.search = defu(options.search, defaultSearchOptions)
```

src/runtime/server/api/indexed-search.ts
farnabaz: Additionally, What do you think about allowing user to define custom query instead of multiple boolean options?

As an example, if we want to search inside a specific directory (let's say `/content/docs`) we can define a custom query in module options and filter contents in side the directory.

src/runtime/composables/search.ts
```diff
@@ -0,0 +1,98 @@
+import MiniSearch, { type Options as MiniSearchOptions } from 'minisearch'
+import { useRuntimeConfig, useFetch } from '#imports'
+
+export const defineMiniSearchOptions = <DataItem>(options: MiniSearchOptions<DataItem>) => {
+  return ref(options)
+}
+
+export const useSearch = async <DataItem>(search: MaybeRefOrGetter<string>, options: MiniSearchOptions<DataItem>) => {
```
Atinux, line 8: Should be `searchContent`

src/runtime/server/api/indexed-search.ts
Atinux: I like this idea yes

Barbapapazes:
Add a warning about this https://github.com/unjs/website/issues/132 in the documentation.

src/module.ts
Barbapapazes: Yes but no. Doing this prevent user to overwrite a field like `fields` which is annoying.

Barbapapazes:
I still need to do some refactor and add a content page!

src/runtime/server/api/indexed-search.ts
Barbapapazes: me too! way better than 100 boolean

src/runtime/composables/search.ts
```diff
@@ -0,0 +1,113 @@
+import MiniSearch, { type Options as MiniSearchOptions } from 'minisearch'
+import { UseFetchOptions } from 'nuxt/app'
```
nobkd, line 2:
```suggestion
import type { UseFetchOptions } from 'nuxt/app'
```

nobkd (changes requested):

Barbapapazes:
Hello @Atinux,

Sorry, I take some times to improve this PR! I've written documentation but I'm open to suggestions.

docs/content/3.guide/2.displaying/3.search.md
```diff
@@ -0,0 +1,82 @@
...
+```js [nuxt.config.js]
+export default defineNuxtConfig({
+  content: {
+    search: true
+  }
+})
+```
+
+::alert{type="info"}
+You can use the search feature with any [Nuxt Content source](../../4.api/3.configuration.md#sources). You can also enable or not the [document-driven feature](../1.writing/7.document-driven.md).
```
nobkd, line 18:
```suggestion
You can use the search feature with any [Nuxt Content source](/api/configuration#sources). You can also enable or not the [document-driven feature](/guide/writing/document-driven).
```

Atinux:
I will wait for @farnabaz deep review since he is the lead of Nuxt Content

src/runtime/composables/search.ts
```diff
@@ -0,0 +1,113 @@
+import MiniSearch, { type Options as MiniSearchOptions } from 'minisearch'
+import type { UseFetchOptions } from 'nuxt/app'
+import { useRuntimeConfig, useFetch } from '#imports'
+
+export const defineMiniSearchOptions = <DataItem>(options: MiniSearchOptions<DataItem>) => {
+  return ref(options)
+}
+
+export const searchContent = async <DataItem>(search: MaybeRefOrGetter<string>, options: { miniSearch?: MaybeRefOrGetter<MiniSearchOptions<DataItem>>, fetch?: MaybeRefOrGetter<UseFetchOptions<string | DataItem[]>> } = {}) => {
```
farnabaz, line 9: What is the use case of `options.fetch`? 
I don't think this is good idea to accept fetch options here. Calling this fetch is more like an internal stuff here.

src/runtime/composables/search.ts
```diff
@@ -0,0 +1,113 @@
...
+    return computed(() => {
+      if (!data.value) {
+        return '[]'
+      }
+
+      const { results } = useIndexedMiniSearch(search, data as Ref<string>, miniSearchOptions!)
+
+      return results.value
+    })
+  }
```
farnabaz, line 30: There are two watchers and one computed inside`useIndexedMiniSearch`, and here we wrapped this function call inside a computed. It seems not right.
I think using `$fetch` and letting user call `searchContent` inside `useAysncData` would be an approach for supporting lazy search.

src/runtime/composables/search.ts
```diff
@@ -0,0 +1,113 @@
...
+    () => toValue(options),
+    () => { indexedMiniSearch.value = createIndexedMiniSearch() },
+    { deep: true }
+  )
+
+  watch(
+    () => toValue(indexedData),
+    () => { indexedMiniSearch.value = createIndexedMiniSearch() },
+    { deep: true }
+  )
```
farnabaz, line 68: Do we need to watch the input data?
The data will not change in production, and it only changes in development. In which we can let HMR do the reloading stuff. WDYT?

src/runtime/composables/search.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  watch(
+    () => toValue(data),
+    (newData) => {
+      miniSearch.value.removeAll()
+      miniSearch.value.addAll(newData)
+    },
+    { deep: true }
+  )
```
farnabaz, line 103: Similar to previous comment, I don't think watching the data is beneficial here

src/runtime/server/api/indexed-search.ts
```diff
@@ -0,0 +1,25 @@
...
+export default defineEventHandler(async (event) => {
+  const runtimeConfig = useRuntimeConfig()
+  const { ignoredTags, ignoreQuery, options } = runtimeConfig.public.content.search
+
+  const files = await serverSearchContent(event, ignoreQuery)
+
+  const sections = (await Promise.all(
+    files
+      .map(page => splitPageIntoSections(page, { ignoredTags }))))
+    .flat()
```
farnabaz, line 15:
```suggestion
  const sections = files
    .map(page => splitPageIntoSections(page, { ignoredTags }))
    .flat()
```

src/runtime/server/api/search.ts
```diff
@@ -0,0 +1,17 @@
...
+export default defineEventHandler(async (event) => {
+  const runtimeConfig = useRuntimeConfig()
+  const { ignoredTags, ignoreQuery } = runtimeConfig.public.content.search
+
+  const files = await serverSearchContent(event, ignoreQuery)
+
+  const sections = (await Promise.all(
+    files
+      .map(page => splitPageIntoSections(page, { ignoredTags }))))
+    .flat()
```
farnabaz, line 14:
```suggestion
  const sections = files
    .map(page => splitPageIntoSections(page, { ignoredTags }))
    .flat()
```

src/runtime/server/api/search.ts
```diff
@@ -0,0 +1,17 @@
+import { defineEventHandler } from 'h3'
+import { serverSearchContent, splitPageIntoSections } from '../search'
+import { useRuntimeConfig } from '#imports'
+
+export default defineEventHandler(async (event) => {
+  const runtimeConfig = useRuntimeConfig()
+  const { ignoredTags, ignoreQuery } = runtimeConfig.public.content.search
+
+  const files = await serverSearchContent(event, ignoreQuery)
```
farnabaz, line 9: I think changing `ignoreQuery` to `filterQuery` will be simpler and more understandable for users.
We can have our default value to filter drafts, partials and ignored contents.

src/module.ts
```diff
@@ -379,6 +432,43 @@ export default defineNuxtModule<ModuleOptions>({
...
+          })
+        } else {
+          nitroConfig.handlers.push({
+            method: 'get',
+            route: nuxt.options.dev
+              ? `${options.api.baseURL}/search.json`
+              : `${options.api.baseURL}/search.${buildIntegrity}.json`,
+            handler: resolveRuntimeModule('./server/api/search')
+          })
+        }
```
farnabaz, line 453: I would merge these two APIs and let behavior controls by runtime config.

src/runtime/composables/search.ts
Barbapapazes: I'll remove it and I think that users usage will tell use!

src/runtime/composables/search.ts
Barbapapazes: I simplify all of this.

Barbapapazes:
Hello,

I refactor a lot using your comments @farnabaz! Hope this is better now!

src/module.ts
farnabaz: WDYT about replacing arrays?
https://github.com/nuxtlabs/studio-module/blob/869f61e2532141c258038fcbafb6c2271636afb9/src/runtime/composables/useStudio.ts#L14-L19

farnabaz:
Thanks @Barbapapazes,
Looks good to me, I'll move the module option under experimental, and we will make it an official feature in v3

PR #2156, opened 2023-07-08
feat: update types

Barbapapazes:
### ❓ Type of change

### 📚 Description

I update types in `useContent` in order to have a better types completion and to allow user to pass it's own type.

### 📝 Checklist


src/runtime/types.d.ts
```diff
@@ -32,30 +32,34 @@ export interface ParsedContentInternalMeta {
...
 
 export interface ParsedContentMeta extends ParsedContentInternalMeta {
   [key: string]: any
 }
 
 export interface ParsedContent extends ParsedContentMeta {
+  /**
+   * Layout
+   */
+  layout: string
```
farnabaz, line 54: `layout` should be an optional field because it may or may not present in front-matter.
I'm also thinking of moving it to `ParsedContentMeta`

src/runtime/composables/content.ts
```diff
@@ -71,12 +71,12 @@ export const useContent = () => {
...
    * Next page from `surround`.
    */
-  const next = computed(() => surround.value?.[1])
+  const next = computed(() => surround.value[1])
 
   /**
    * Previous page from `surround`.
    */
-  const prev = computed(() => surround.value?.[0])
+  const prev = computed(() => surround.value[0])
```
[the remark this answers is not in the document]
Barbapapazes: In which case? As I can experiment, there is always a value and the object could be `[null, null]`

src/runtime/types.d.ts
Barbapapazes: Yes, good catch for the undefined.

Should we moved it to ParsedMeta, I don't know. I myself asked the question and I understand that ParsedMeta are calculated and because layout is not, I added here. But i'm completely open to move it.

farnabaz: I think `ParsedMeta` is better place for it

src/runtime/composables/content.ts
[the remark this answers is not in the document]
farnabaz: `surrounds` is set inside `documentDriven.ts` only if HTTP call succeeds, if HTTP call fails surround will be `null`

Also, it will be null if users disable `surrounds` in document driven.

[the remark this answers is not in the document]
Barbapapazes: Ho yes, that's a good point.

manniL:
Will this allow typing the `page` that comes back by any chance?

Barbapapazes:
> Will this allow typing the `page` that comes back by any chance? 

Could you tell me which `pages`?

manniL:
@Barbapapazes I was thinking of sth like:

```ts
const { page } = useContent<MyType>()
      // ^? MyType
```

Barbapapazes:
> @Barbapapazes I was thinking of sth like:
> 
> ```ts
> const { page } = useContent<MyType>()
>       // ^? MyType
> ```
> 
> This already works fine for `queryContent`, e.g. via `queryContent<Article>('/articles')`.

Yes, that's will be possible!

src/runtime/composables/content.ts
```diff
@@ -71,12 +71,12 @@ export const useContent = () => {
   /**
    * Next page from `surround`.
    */
-  const next = computed(() => surround.value?.[1])
+  const next = computed(() => surround?.value[1])
```
farnabaz, line 74:
```suggestion
  const next = computed(() => surround.value?.[1])
```
Surround is a ref, and we should check its value for `undefined`

src/runtime/composables/content.ts
```diff
@@ -71,12 +71,12 @@ export const useContent = () => {
...
 
   /**
    * Previous page from `surround`.
    */
-  const prev = computed(() => surround.value?.[0])
+  const prev = computed(() => surround?.value[0])
```
farnabaz, line 79:
```suggestion
  const prev = computed(() => surround.value?.[0])
```

surround is a ref, and we should check its value for `undefined`

farnabaz (approved):
Thanks ❤️

PR #2157, opened 2023-07-08
fix: remove favicon handling

Barbapapazes:
### ❓ Type of change

### 📚 Description

I remove no more needed code since this (https://github.com/nuxt/framework/pull/5274#issuecomment-1212512771) is resolved.

### 📝 Checklist


Atinux:
The issue is actually when we don't have any `public/favicon.ico`, the browser will fetch it automatically.

farnabaz (approved):
I can confirm that nitro return empty gif image in case of missing favicon

https://github.com/unjs/nitro/blob/1b9f51b7955d2490b1fd8da73af40780f30a1235/src/runtime/renderer.ts#L18-L26

PR #2169, opened 2023-07-12
fix: allow ] in filename code block

### LINKED ISSUE #2166: in documents, filename "[...slug].vue" above code block display not correctly
### Environment

- Operating System: Darwin
- Node Version:     v16.18.0
- Nuxt Version:     0.0.0
- Nitro Version:    -
- Package Manager:  PNPM
- Builder:          webpack
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://content.nuxtjs.org/api/components/content-doc#contentdoc

### Describe the bug

image

### Additional context

_No response_

### Logs

_No response_

Barbapapazes:
### ❓ Type of change

### 📚 Description

fix #2166 

For a code block, if filename was `[...page].vue`, regex was stoping at `[...page`. This PR fix it.

### 📝 Checklist

### DIFF
--- a/docs/content/4.api/1.components/1.content-doc.md
+++ b/docs/content/4.api/1.components/1.content-doc.md
@@ -73,7 +73,7 @@ The `empty`{lang=ts} slot can be used to display a default content before render
 
 ### Default
 
-```html [pages/[...slug.vue\\]]
+```html [pages/[...slug\\].vue]
 <template>
   <main>
     <!-- Similar to <ContentDoc :path="$route.path" /> -->
--- a/docs/content/4.api/1.components/1.content-doc.md
+++ b/docs/content/4.api/1.components/1.content-doc.md
@@ -73,7 +73,7 @@ The `empty`{lang=ts} slot can be used to display a default content before render
 
 ### Default
 
-```html [pages/[...slug.vue]]
+```html [pages/[...slug.vue\\]]
 <template>
   <main>
     <!-- Similar to <ContentDoc :path="$route.path" /> -->
--- a/docs/content/4.api/1.components/2.content-renderer.md
+++ b/docs/content/4.api/1.components/2.content-renderer.md
@@ -27,7 +27,7 @@ Other types will currently be passed to default slot via `v-slot="{ data }"` or 
 
 The `default`{lang=ts} slot can be used to render the content via `v-slot="{ data }"` syntax.
 
-```html [pages/[...slug.vue]]
+```html [pages/[...slug.vue\\]]
 <script setup lang="ts">
 const { data } = await useAsyncData('page-data', () => queryContent('/hello').findOne())
 </script>
@@ -44,7 +44,7 @@ const { data } = await useAsyncData('page-data', () => queryContent('/hello').fi
 
 The `empty`{lang=ts} slot can be used to display a default content when the document is empty:
 
-```html [pages/[...slug.vue]]
+```html [pages/[...slug.vue\\]]
 <script setup lang="ts">
 const { data } = await useAsyncData('page-data', () => queryContent('/hello').findOne())
 </script>
@@ -63,7 +63,7 @@ const { data } = await useAsyncData('page-data', () => queryContent('/hello').fi
 ::alert{type="info"}
 Note that when you use default slot and `<ContentRendererMarkdown>` in your template you need to pass `components` to `<ContentRendererMarkdown>`.
 
-```html [pages/[...slug].vue]
+```html [pages/[...slug\\].vue]
 <script setup lang="ts">
 const { data } = await useAsyncData('page-data', () => queryContent('/hello').findOne())
 
--- a/docs/content/4.api/1.components/2.content-renderer.md
+++ b/docs/content/4.api/1.components/2.content-renderer.md
@@ -27,7 +27,7 @@ Other types will currently be passed to default slot via `v-slot="{ data }"` or 
 
 The `default`{lang=ts} slot can be used to render the content via `v-slot="{ data }"` syntax.
 
-```html [pages/[...slug.vue\\]]
+```html [pages/[...slug\\].vue]
 <script setup lang="ts">
 const { data } = await useAsyncData('page-data', () => queryContent('/hello').findOne())
 </script>
@@ -44,7 +44,7 @@ const { data } = await useAsyncData('page-data', () => queryContent('/hello').fi
 
 The `empty`{lang=ts} slot can be used to display a default content when the document is empty:
 
-```html [pages/[...slug.vue\\]]
+```html [pages/[...slug\\].vue]
 <script setup lang="ts">
 const { data } = await useAsyncData('page-data', () => queryContent('/hello').findOne())
 </script>
--- a/docs/content/4.api/1.components/4.content-query.md
+++ b/docs/content/4.api/1.components/4.content-query.md
@@ -39,7 +39,7 @@ The `<ContentQuery>`{lang=html} component fetches a document and gives access to
 
 The `default`{lang=ts} slot can be used to render the content via `v-slot="{ data }"`{lang=html} syntax.
 
-```html [pages/[...slug.vue]]
+```html [pages/[...slug.vue\\]]
 <!-- Similar to <ContentDoc :path="$route.path" /> -->
 <template>
   <main>
@@ -63,7 +63,7 @@ The `not-found`{lang=ts} slot can be used to display a default content before re
 <template>
   <main>
     <ContentQuery path="/about/authors" :where="{ type: 'csv' }">
-      <template #default="{ data }"> 
+      <template #default="{ data }">
         <ul>
           <li v-for="author of data" :key="author.name">
             {{ author.name }}
--- a/docs/content/4.api/1.components/4.content-query.md
+++ b/docs/content/4.api/1.components/4.content-query.md
@@ -39,7 +39,7 @@ The `<ContentQuery>`{lang=html} component fetches a document and gives access to
 
 The `default`{lang=ts} slot can be used to render the content via `v-slot="{ data }"`{lang=html} syntax.
 
-```html [pages/[...slug.vue\\]]
+```html [pages/[...slug\\].vue]
 <!-- Similar to <ContentDoc :path="$route.path" /> -->
 <template>
   <main>
--- a/docs/content/4.api/1.components/7.prose.md
+++ b/docs/content/4.api/1.components/7.prose.md
@@ -71,9 +71,9 @@ Component properties will be:
 Check out the [highlight options](/api/configuration#highlight) for more about the syntax highlighting.
 
 ::alert{type="warning"}
-One limitation is you can't put `[` or `]` in the meta info. Otherwise, it will break the parsing.
+If you want to use `]` in the filename, you need to escape it with 2 backslashes: `\\]`. This is necessary since JS will automatically escape the backslash in a string so `\\]` will be resolved as `\]`.
 
-This is due to how code block header to allow `[` and `]` in the filename.
+Internally, a regex is used to extract filename.
 ::
 
 ## `ProseCodeInline`
--- a/docs/content/4.api/1.components/7.prose.md
+++ b/docs/content/4.api/1.components/7.prose.md
@@ -70,6 +70,12 @@ Component properties will be:
 
 Check out the [highlight options](/api/configuration#highlight) for more about the syntax highlighting.
 
+::alert{type="warning"}
+One limitation is you can't put `[` or `]` in the meta info. Otherwise, it will break the parsing.
+
+This is due to how code block header to allow `[` and `]` in the filename.
+::
+
 ## `ProseCodeInline`
 
 [:icon{name="fa-brands:github" class="inline -mt-1 w-6"} Source](https://github.com/nuxt/content/tree/main/src/runtime/components/Prose/ProseCodeInline.vue)
--- a/docs/content/4.api/1.components/7.prose.md
+++ b/docs/content/4.api/1.components/7.prose.md
@@ -71,9 +71,7 @@ Component properties will be:
 Check out the [highlight options](/api/configuration#highlight) for more about the syntax highlighting.
 
 ::alert{type="warning"}
-If you want to use `]` in the filename, you need to escape it with 2 backslashes: `\\]`. This is necessary since JS will automatically escape the backslash in a string so `\\]` will be resolved as `\]`.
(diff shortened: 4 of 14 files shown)


src/runtime/markdown-parser/handler/utils.ts
```diff
@@ -20,7 +20,7 @@ export function parseThematicBlock (lang: string) {
 
   const language = lang.replace(/[{|[](.+)/, '').match(/^[^ \t]+(?=[ \t]|$)/)
   const highlightTokens = lang.match(/{([^}]*)}/)
-  const filenameTokens = lang.match(/\[([^\]]*)\]/)
+  const filenameTokens = lang.match(/\[(.*)\]/) /** Allow ']' in filename. @see https://regex101.com/r/Hm0t5m/1 */
```
nobkd, line 23: I think we used `[^\]]` to allow using lists in the meta part.
Here's an example what would not work anymore if we match the longest string.
> \```language [filename]{hightlights} meta=[...]

It would match the content `filename]{hightlights} meta=[...`

nobkd: I thought we could use something like `\[` in the markdown and prevent matching those somehow

Barbapapazes: Oooh I understand. I do it too quickly!

nobkd: I think, that this might work: `/\[((\\]|[^\]])*)\]/`

---

This part `(\\]|[^\]])*` matches one or a sequence of: `\]` or characters that are not `]`

Only if our string has no more `]` after a `\]` it terminates the match at a `\]`.

e.g.

```
```ts [[...slug\].vue] meta=[]
```

=> `[...slug\].vue` as the first group match. Then we only have to replace `\]` which in reality is `\\]` with `]`-es

nobkd, line 23:
```suggestion
  const filenameTokens = lang.match(/\[((\\]|[^\]])*)\]/)
```

nobkd: And because of GitHub, I'm not allowed to suggest changes out of some range:

Change for L29:

```diff
-    filename: Array.isArray(filenameTokens) && filenameTokens[1] ? filenameTokens[1] : undefined,
+    filename: Array.isArray(filenameTokens) && filenameTokens[1] ? filenameTokens[1].replace(/\\]/g, ']') : undefined,
```

src/runtime/markdown-parser/handler/utils.ts
```diff
@@ -20,7 +20,7 @@ export function parseThematicBlock (lang: string) {
...
   const meta = lang.replace(/^\w*\s*(\[[^\]]*\]|\{[^}]*\})?\s*(\[[^\]]*\]|\{[^}]*\})?\s*/, '')
```
nobkd, line 24:
```suggestion
  const meta = lang.replace(/^\w*\s*(\[((\\]|[^\]])*)\]|\{[^}]*\})?\s*(\[((\\]|[^\]])*)\]|\{[^}]*\})?\s*/, '')
```

To replace the correct part to have the meta left

nobkd (changes requested):
This should be all the necessary changes, but we'll see..
This is currently not really tested.

I tried both `\\]` and `\]` and it seems that we need the first variant with two backslashes to get one real backslash from the Markdown.

So this should work:
```md
```ts [[...slug\\].vue] ...
```

src/runtime/markdown-parser/handler/utils.ts
nobkd: Question is... is `\` or `\\` in Markdown a backslash?
Do we have to write `\\]` or `\]` in the markdown file?

Barbapapazes: Yes but if there is no meta, this regex is not working

https://regex101.com/r/bG0Net/1

Barbapapazes: For test, I think that we need to extract these utilities into simple function to be able to unit tests them.

Barbapapazes:
I will also update documentation to add to warning about usage.

src/runtime/markdown-parser/handler/utils.ts
Barbapapazes: I refactor this to be more understandable.

src/runtime/markdown-parser/handler/utils.ts
```diff
@@ -18,15 +18,19 @@ export function parseThematicBlock (lang: string) {
     }
   }
 
-  const language = lang.replace(/[{|[](.+)/, '').match(/^[^ \t]+(?=[ \t]|$)/)
-  const highlightTokens = lang.match(/{([^}]*)}/)
-  const filenameTokens = lang.match(/\[([^\]]*)\]/)
-  const meta = lang.replace(/^\w*\s*(\[[^\]]*\]|\{[^}]*\})?\s*(\[[^\]]*\]|\{[^}]*\})?\s*/, '')
+  const languageMatches = lang.replace(/[{|[](.+)/, '').match(/^[^ \t]+(?=[ \t]|$)/)
+  const highlightTokensMatches = lang.match(/{([^}]*)}/)
+  const filenameMatches = lang.match(/\[((\S)*)\]/) /** Allow ']' in filename. */
```
nobkd, line 23: I can see that you don't want an escape character to have `]` in your filename, but we didn't want to depend on spaces. If we now change this, it could be a breaking change.

My variant from before would introduce an escape character (e.g. `\\`) for `]` but
- would not need spaces after the last `]` that belongs to the filename
- allow spaces in the filename.


You could not do this anymore:

```md
```ts [filename]meta=[]
```

or

```md
```ts [filename/title with spaces]
```

Spaces in the filename for example are already used in the `nuxt/nuxt` docs. Some examples:
- https://github.com/nuxt/nuxt/blob/350788419c414317354de9900cd049fa750456cc/docs/7.migration/4.meta.md?plain=1#L27
- https://github.com/nuxt/nuxt/blob/350788419c414317354de9900cd049fa750456cc/docs/1.getting-started/5.routing.md?plain=1#L17
- https://github.com/nuxt/nuxt/blob/350788419c414317354de9900cd049fa750456cc/docs/1.getting-started/5.routing.md?plain=1#L25

---

#2067 was the last change to this regexes, there should also be some ideas.

nobkd (changes requested):

src/runtime/markdown-parser/handler/utils.ts
Barbapapazes: you've convinced me. I will update regex to allow esacape ] and update docs in that ways (documenting why we choose this method).

nobkd, line 23:
```suggestion
  const filenameMatches = lang.match(\[((\\]|[^\]])*)\]/)
```

src/runtime/markdown-parser/handler/utils.ts
```diff
@@ -18,15 +18,19 @@ export function parseThematicBlock (lang: string) {
...
+    .replace(highlightTokensMatches?.[0] ?? '', '')
+    .replace(filenameMatches?.[0] ?? '', '')
+    .trim()
 
   return {
-    language: language ? language[0] : undefined,
-    highlights: parseHighlightedLines(highlightTokens && highlightTokens[1]),
-    filename: Array.isArray(filenameTokens) && filenameTokens[1] ? filenameTokens[1] : undefined,
+    language: languageMatches?.[0] || undefined,
+    highlights: parseHighlightedLines(highlightTokensMatches?.[1] || undefined),
```
nobkd, line 32:
```suggestion
    highlights: parseHighlightedLines(highlightTokensMatches?.[1].replace(/\\]/g, ']') || undefined),
```

nobkd (changes requested):

src/runtime/markdown-parser/handler/utils.ts
```diff
@@ -18,15 +18,19 @@ export function parseThematicBlock (lang: string) {
...
-  const filenameTokens = lang.match(/\[([^\]]*)\]/)
-  const meta = lang.replace(/^\w*\s*(\[[^\]]*\]|\{[^}]*\})?\s*(\[[^\]]*\]|\{[^}]*\})?\s*/, '')
+  const languageMatches = lang.replace(/[{|[](.+)/, '').match(/^[^ \t]+(?=[ \t]|$)/)
+  const highlightTokensMatches = lang.match(/{([^}]*)}/)
+  const filenameMatches = lang.match(/\[((\S)*)\]/) /** Allow ']' in filename. */
+  const meta = lang
+    .replace(languageMatches?.[0] ?? '', '')
+    .replace(highlightTokensMatches?.[0] ?? '', '')
+    .replace(filenameMatches?.[0] ?? '', '')
+    .trim()
```
nobkd, line 28: Thanks for this, it makes this part so much more readable! ❤️

Barbapapazes:
@nobkd we have an issue because a JS string is automatically escape `\` in string. So `[[...page\].vue]` is resolved as `[[...page].vue]` 

image

Barbapapazes:
We will need to use a double `\`. I'll update PR with this usage.

nobkd:
See this https://github.com/nuxt/content/pull/2169#pullrequestreview-1527240638

We need to use double escaping: `\\]`

There's probably a better escape character, but `\` is the generally used one, so I went with it. Do you have a better idea how to escape?

nobkd:
> We will need to use a double `\`. I'll update PR with this usage.

Exactly

Barbapapazes:
> See this [#2169 (review)](https://github.com/nuxt/content/pull/2169#pullrequestreview-1527240638)
> 
> We need to use double escaping: `\\]`
> 
> There's probably a better escape character, but `\` is the generally used one, so I went with it. Do you have a better idea how to escape?

Oh yes, I've not seen this comment! Thanks

docs/content/4.api/1.components/2.content-renderer.md
```diff
@@ -27,7 +27,7 @@ Other types will currently be passed to default slot via `v-slot="{ data }"` or
 
 The `default`{lang=ts} slot can be used to render the content via `v-slot="{ data }"` syntax.
 
-```html [pages/[...slug.vue]]
+```html [pages/[...slug.vue\\]]
```
nobkd, line 30:
```suggestion
```html [pages/[...slug\\].vue]
```

docs/content/4.api/1.components/2.content-renderer.md
```diff
@@ -44,7 +44,7 @@ const { data } = await useAsyncData('page-data', () => queryContent('/hello').fi
 
 The `empty`{lang=ts} slot can be used to display a default content when the document is empty:
 
-```html [pages/[...slug.vue]]
+```html [pages/[...slug.vue\\]]
```
nobkd, line 47:
```suggestion
```html [pages/[...slug\\].vue]
```

docs/content/4.api/1.components/4.content-query.md
```diff
@@ -39,7 +39,7 @@ The `<ContentQuery>`{lang=html} component fetches a document and gives access to
 
 The `default`{lang=ts} slot can be used to render the content via `v-slot="{ data }"`{lang=html} syntax.
 
-```html [pages/[...slug.vue]]
+```html [pages/[...slug.vue\\]]
```
nobkd, line 42:
```suggestion
```html [pages/[...slug\\].vue]
```

docs/content/4.api/1.components/7.prose.md
```diff
@@ -70,6 +70,12 @@ Component properties will be:
 
 Check out the [highlight options](/api/configuration#highlight) for more about the syntax highlighting.
 
+::alert{type="warning"}
+If you want to use `]` in the filename, you need to escape it with 2 backslashes: `\\]`. This is necessary since JS will automatically escape the backslash in a string so `\\]` will be resolved as `\]`.
+
+Internally, a regex is used to extract filename.
```
nobkd:
```suggestion
Internally, a regex is used to extract the filename.
```

docs/content/4.api/1.components/1.content-doc.md
```diff
@@ -73,7 +73,7 @@ The `empty`{lang=ts} slot can be used to display a default content before render
 
 ### Default
 
-```html [pages/[...slug.vue]]
+```html [pages/[...slug.vue\\]]
```
nobkd, line 76:
```suggestion
```html [pages/[...slug\\].vue]
```

docs/content/4.api/1.components/7.prose.md
```diff
@@ -70,6 +70,10 @@ Component properties will be:
 
 Check out the [highlight options](/api/configuration#highlight) for more about the syntax highlighting.
 
+::alert{type="warning"}
+If you want to use `]` in the filename, you need to escape it with 2 backslashes: `\\]`. This is necessary since JS will automatically escape the backslash in a string so `\]` will be resolved as `]` breaking our regex.
```
nobkd, line 74: Is the rendered Markdown for `\]` and `\\] ` properly rendered, so we see the backslash(es)?
I think in code blocks we don't have to escape, but I'm not sure...

nobkd (changes requested):
Here some small fixes, as in #2164

Everything else looks great to me, thanks 👍

playground/basic/components/content/ProseCode.vue
```diff
@@ -0,0 +1,38 @@
+<script lang="ts" setup>
+const props = defineProps<{
+    code: string,
+    language?: string,
+    filename?: string,
+    highlights: number[],
+    meta?: string
+  }>()
+
+console.log(props)
```
farnabaz, line 10:
```suggestion
```

playground/basic/components/content/ProseCode.vue
```diff
@@ -0,0 +1,36 @@
+<script lang="ts" setup>
+const props = defineProps<{
```
farnabaz, line 2:
```suggestion
defineProps<{
```

farnabaz (approved):
Thanks guys.
You're awesome ❤️

nobkd:
Just to have this in one place, I think this bug was previously introduced with #2067

https://github.com/nuxt/content/pull/2067/files#diff-fd60d817a96ebccaa6b1456127b1e4589b3e10487ddb84362ab6e41ab1ede143L22-R23

I think, there was decided, that we could allow `[]` and `{}` in the meta part, and therefore changed the regex...

PR #2167, opened 2023-07-12
docs: use lowercase typings, use `|` instead of `or` in type declaration

nobkd:
### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

Other declarations follow the same schema.

PR #2170, opened 2023-07-12
fix: uppercase in path

### LINKED ISSUE #2162: Enable use hyphen in path folder

Currently, using hyphens in the path folder causes routing errors (e.g. /content/zh-CN/xxx).
I wish I could use it.

Jannchie:
### 🔗 Linked issue

close #2162 

### ❓ Type of change

### 📚 Description

Currently, if we construct a path like /zh-CN/foo.md, the corresponding document cannot be found. This is because the path is automatically converted to lowercase. I think the existence of such a path is reasonable and should not be converted. I don't understand why this is done, so in order not to introduce destructive changes, I added a flag named keepUppercase to allow uppercase letters in the path.

I wasn't sure where this flag should be placed, so I tentatively put it under the experimental options.

Resolves #2162 

### 📝 Checklist

### DIFF
--- a/docs/content/4.api/3.configuration.md
+++ b/docs/content/4.api/3.configuration.md
@@ -416,4 +416,4 @@ Toggles the document-driven mode.
 - Type: `Boolean`{lang=ts}
 - Default: `false`{lang=ts}
 
-Whether to respect the case of the file path when generating the route. Defaults to `false`, which means the route will be generated in lowercase. If set to `true`, the route will be generated with the same case as the file path.
+Whether to respect the case of the file path when generating the route. Defaults to `false`, which means the route will be generated in lowercase. For example, `content/en-US/foo.md` will resolve to the `en-us/foo` path. This may not be what you expect.If set to `true`, the route will be generated with the same case as the file path. `content/en-US/foo.md` will resolve to the `en-US/foo` path.
--- a/docs/content/4.api/3.configuration.md
+++ b/docs/content/4.api/3.configuration.md
@@ -411,9 +411,9 @@ Toggles the document-driven mode.
 | `injectPage`      |      `boolean`      | Inject `[...slug].vue` pre-configured page                     |
 | `trailingSlash`   |      `boolean`      | Add a slash at the end of  `canonical` and `og:url`            |
 
-## `keepUppercase`
+## `respectPathCase`
 
 - Type: `Boolean`{lang=ts}
 - Default: `false`{lang=ts}
 
-Whether to keep uppercase characters in the generated routes.
+Whether to respect the case of the file path when generating the route. Defaults to `false`, which means the route will be generated in lowercase. If set to `true`, the route will be generated with the same case as the file path.
--- a/docs/content/4.api/3.configuration.md
+++ b/docs/content/4.api/3.configuration.md
@@ -410,3 +410,10 @@ Toggles the document-driven mode.
 | `layoutFallbacks` |     `string[]`      | A list of `globals` key to use to find a layout fallback.      |
 | `injectPage`      |      `boolean`      | Inject `[...slug].vue` pre-configured page                     |
 | `trailingSlash`   |      `boolean`      | Add a slash at the end of  `canonical` and `og:url`            |
+
+## `keepUppercase`
+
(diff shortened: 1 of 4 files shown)


nobkd:
Can you add this option to the docs too?
I think this is the right place: https://github.com/nuxt/content/blob/main/docs/content/4.api/3.configuration.md

Also, I don't really think, that this is a breaking change. So I think that it does not have to be in the experimental features. What do you think?

Jannchie:
> Also, I don't really think, that this is a breaking change. So I think that it does not have to be in the experimental features. What do you think?

You are right, I moved the location of this option and added the documentation! Thanks for the continued follow up.

src/module.ts
```diff
@@ -220,6 +220,12 @@ export interface ModuleOptions {
     injectPage?: boolean
     trailingSlash?: boolean
   },
+  /**
+   * Enable to keep uppercase characters in the generated routes.
+   *
+   * @default false
+   */
+  keepUppercase: boolean
```
Barbapapazes, line 228: This name may be too vague since we do not understand it's related to path. 

Maybe something like respectPathCase, when false, it lower case or respectFilenameCase

Jannchie: I've rename the name to `respectPathCase`.

docs/content/4.api/3.configuration.md
```diff
@@ -410,3 +410,10 @@ Toggles the document-driven mode.
 | `layoutFallbacks` |     `string[]`      | A list of `globals` key to use to find a layout fallback.      |
 | `injectPage`      |      `boolean`      | Inject `[...slug].vue` pre-configured page                     |
 | `trailingSlash`   |      `boolean`      | Add a slash at the end of  `canonical` and `og:url`            |
+
+## `respectPathCase`
+
+- Type: `Boolean`{lang=ts}
```
nobkd, line 416:
```suggestion
- Type: `boolean`{lang=ts}
```
Currently making types in the docs lowercase. See #2167

nobkd (changes requested):

docs/content/4.api/3.configuration.md
Jannchie: fixed..!

docs/content/4.api/3.configuration.md
```diff
@@ -410,3 +410,10 @@ Toggles the document-driven mode.
 | `layoutFallbacks` |     `string[]`      | A list of `globals` key to use to find a layout fallback.      |
 | `injectPage`      |      `boolean`      | Inject `[...slug].vue` pre-configured page                     |
 | `trailingSlash`   |      `boolean`      | Add a slash at the end of  `canonical` and `og:url`            |
+
+## `respectPathCase`
```
Barbapapazes, line 414: Maybe you could provide an example to help dev to understand types of issues that this options can solve

Barbapapazes:
Nice PR!

docs/content/4.api/3.configuration.md
Jannchie: Added!

Barbapapazes: Perfect 👌

farnabaz (approved):
Perfect 👍 
Thanks ❤️

src/runtime/transformers/path-meta.ts
```diff
@@ -67,7 +67,8 @@ const isPartial = (path: string): boolean => path.split(/[:/]/).some(part => par
  * @returns generated slug
  */
 export const generatePath = (path: string, { forceLeadingSlash = true } = {}): string => {
-  path = path.split('/').map(part => slugify(refineUrlPart(part), { lower: true })).join('/')
+  const { content } = useRuntimeConfig().public
```
farnabaz, line 70: We should avoid using Nuxt generated helpers (like `useRuntimeConfig`) in transformers. Transformers and in general the content parser could be able to work outside of Nuxt Context.

There is already several usecases of parser outside of Nuxt.

`respectPathCase`  could be passed in second argumant of this function and transformer itseld

farnabaz (changes requested):

farnabaz (approved):

PR #2221, opened 2023-08-08
feat: support sync components

Atinux:
### ❓ Type of change

### 📚 Description

Add the `components` options to specify the components to add in the main bundle, useful for non-global components from UI library or a way to optimize global components that we use in many pages.

### 📝 Checklist

### DIFF
--- a/docs/content/4.api/3.configuration.md
+++ b/docs/content/4.api/3.configuration.md
@@ -35,7 +35,7 @@ export default defineNuxtConfig({
 
 ## `components`
 
-- Type: `array`{lang=ts}
+- Type: `string[]`{lang=ts}
 - Default: `[]`{lang=ts}
 
 Specify the list of components which are not in `components/content/` or not defined as global component to add them in the main bundle (included in all pages). Useful for components used in UI library.
--- a/docs/content/4.api/3.configuration.md
+++ b/docs/content/4.api/3.configuration.md
@@ -33,6 +33,21 @@ export default defineNuxtConfig({
 })
 ```
 
+## `components`
+
+- Type: `array`{lang=ts}
+- Default: `[]`{lang=ts}
+
+Specify the list of components which are not in `components/content/` or not defined as global component to add them in the main bundle (included in all pages). Useful for components used in UI library.
+
+```ts [nuxt.config.ts]
+export default defineNuxtConfig({
+  content: {
+    components: ['UButton']
+  }
+})
+```
+
 ## `watch`
 
 - Type: `object | false`{lang=ts}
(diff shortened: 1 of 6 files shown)


src/module.ts
```diff
@@ -58,6 +60,12 @@ export interface ModuleOptions {
      */
     baseURL: string
   }
+  /**
+   * List the components that will be used in markdown.
+   *
+   * @default []
+   */
+  components: Array<string>
```
nobkd, line 68:
```suggestion
  components: string[]
```

docs/content/4.api/3.configuration.md
```diff
@@ -33,6 +33,21 @@ export default defineNuxtConfig({
 })
 ```
 
+## `components`
+
+- Type: `array`{lang=ts}
```
nobkd, line 38:
```suggestion
- Type: `string[]`{lang=ts}
```

nobkd (changes requested):

src/module.ts
```diff
@@ -610,6 +619,21 @@ export default defineNuxtModule<ModuleOptions>({
     // @ts-ignore
     await nuxt.callHook('content:context', contentContext)
 
+    // Add sync components to use not global components
+    const components = (options.components || []).map(pascalCase)
+
+    addPluginTemplate({
```
farnabaz, line 625: It would be better if we mark these components as global instead of registering a custom plugin. Marking them as global will help other modules to be list these components as global ones. Modules like `nuxt-component-meta`

Atinux: This feature is to avoid using global components and put them in the main bundle of the application actually. I want to see if this improves the performance

farnabaz: I see, Nuxt registers lazy version of global components, but here the goal is to include it directly inside main bundle.

https://github.com/nuxt/nuxt/blob/ce9aecca4a39e1d5e36e6b92c77e706f8f577445/packages/nuxt/src/components/templates.ts#L50-L63

src/module.ts
```diff
@@ -610,6 +619,21 @@ export default defineNuxtModule<ModuleOptions>({
...
+      filename: 'plugins/content-components.ts',
+      getContents: () => {
+        return `${genImport('#components', components)}
```
farnabaz, line 628:
```suggestion
        return `import { defineNuxtPlugin } from '#app/nuxt'
${genImport('#components', components)}
```

danielroe:
Merged https://github.com/nuxt/nuxt/pull/22558 to Nuxt - might be worth experimenting with but do note my comments in the PR. It might be worth iterating on a few possible alternatives within content.

PR #2225, opened 2023-08-11
refactor: use vite ws connection

### LINKED ISSUE #1914: watch config option : allow 0.0.0.0 mask

I'm using nuxt content in a docker. Hence the ws URL is not `localhost:4000` but `172.19.0.2:4000`, which I can't connect to if the websocket network mask is not set to `0.0.0.0`.

Add a `ws.mask` option in nuxt content config.

Having a `ws.url = 172.19.0.2:4000` option would also work but would be less comfortable to use and less generic.

For now, I simply have to disable the `watch` option.

Cheers

Barbapapazes:
### 🔗 Linked issue

fix #1914

### ❓ Type of change

### 📚 Description

Use the web socket connection from the vite server.

_I still need to clean a lot of thing but I think that it's a great start_.

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "@nuxtjs/eslint-config-typescript": "latest",
     "@types/ws": "^8.5.5",
     "c8": "^8.0.1",
+    "chokidar": "^3.5.3",
     "csvtojson": "^2.0.10",
     "eslint": "^8.46.0",
     "globby": "^13.2.2",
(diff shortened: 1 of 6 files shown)


Barbapapazes:
Works fine but I got a double log shown (so a double request is made).

Barbapapazes:
Maybe, there is some build optimization to do (to avoid having chokidar?).

farnabaz:
Thanks for the PR @Barbapapazes,

Nuxt also supports Webpack, So we also need to find/implement a solution for webpack.

More importantly, In the beta version of Content@v2 I've tried to use build tools dev sockets and drop Content's custom server. However, the work left unmerged. Seems this approach is not aligned with framework's guidelines.

/cc @danielroe

PR #1982, opened 2023-03-28
docs: format docs & update links & fix mistakes

nobkd:
### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Tried to standardize Markdown docs
Updated links to nuxt.com
Fixed some mistakes

**Added To-dos to `guide/migration/from-v1` where docs don't look quite right, but I'm not sure how to correct it**

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

PR #2228, opened 2023-08-15
docs(mdc): fix {attribute} examples

nobkd:
### ❓ Type of change

### 📚 Description

As Tailwind isn't used by Docus anymore, the class names aren't providing styling by themselves. I added the corresponding style tags to fix this.
This has also been [changed for lines 347-349](https://github.com/nuxt/content/commit/1823c0defcc836ebfdfc0f1bf9be1e8ac19f0cc7#diff-486c5f2e2433a70cd8bace1d6253fd758d7ef340bb21fa57dcf135ec9eca2801R347-R349) in a previous PR #1982.

I am not sure, if it is good to do this, because the class name styling won't work if someone tries doing it with Docus.
Maybe a warning or info should be added, that the styling with classes just works with e.g. Tailwind?
I mean... the props are assigned, just the styling would be missing.

### 📝 Checklist

### DIFF
--- a/docs/content/3.guide/1.writing/3.mdc.md
+++ b/docs/content/3.guide/1.writing/3.mdc.md
@@ -320,6 +320,8 @@ To create inline spans in your text you can use the `[]` identifier.
 
 Attributes are useful for highlighting and modifying part of paragraph. The syntax is nearly similar to inline components and markdown links syntax.
 
+Possible values ​​are all named attributes, classes with the notation `.class-name` and an ID with `#id-name`.
+
 ::code-group
   ```md [Code]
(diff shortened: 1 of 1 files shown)


farnabaz:
The section is about the inline attributes, so I think it would also make sense that in the code sample, we use `style` attribute to show the usage, it is not necessary to use classes.


```
  ```md [Code]
  Hello [World]{style="background-color: var(--color-primary-500)" .custom-class}!
  \```

  ::code-block{label="Preview"}
  Hello [World]{style="background-color: var(--color-primary-500)" .custom-class}!
  ::
```

farnabaz (approved):

PR #2298, opened 2023-09-07
fix(module): remove deprecated `resolveModule`

farnabaz:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -2,7 +2,6 @@ import fs from 'fs'
 import {
   addPlugin,
   defineNuxtModule,
-  resolveModule,
   createResolver,
   addImports,
   addComponentsDir,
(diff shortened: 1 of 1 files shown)


src/module.ts
```diff
@@ -307,7 +306,7 @@ export default defineNuxtModule<ModuleOptions>({
   },
   async setup (options, nuxt) {
     const { resolve } = createResolver(import.meta.url)
-    const resolveRuntimeModule = (path: string) => resolveModule(path, { paths: resolve('./runtime') })
+    const resolveRuntimeModule = (path: string) => resolve('./runtime', path)
```
pi0, line 309: Not that resolve just joint paths. It doesn't check extension. You might want to use resolver.resolePath  but have to try

farnabaz: AFAIK almost in every cases of our usages Nuxt takes care of the extension matching and module works fine with `resolve` function

pi0: If you want to rely on that (and it works with published module) surely!

farnabaz: I've tested the build in a real project and it's fine. Will merge and check on edge channel too.
Thanks

Issue #2469, opened 2023-12-19
not compatible with TypeScript `Bundler` module resolution


danielroe:
We are planning to enable `Bundler` module resolution by default in future (see https://github.com/nuxt/nuxt/pull/22821).

Nuxt Content is currently not compatible. Here are some issues I think need to be addressed:
- [ ] subpath export for runtime utilities/types that are intended to be directly imported. (Currently it's documented to use `@nuxt/content/dist/*` - we should probably add this as a subpath _at a minimum_ and then recommend using a more ergonomic path
- [ ] emit `.d.mts` and `.mjs` (with `.d.ts` as legacy fallback) within runtime directory - we are currently _missing_ the `.mjs` files which also leads to broken import types. I think this likely needs to be handled by default by `@nuxt/module-builder`

PR #2470, opened 2023-12-19
fix: ensure `bundler` module resolution works with runtime type imports

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/content/issues/2469

### ❓ Type of change

### 📚 Description

This adds support for Bundler TypeScript module resolution. It changes extensions of runtime dir files to `.js` which should be safe as this is a `type: module` project.

I also add a subpath for `@nuxt/content/dist/runtime/types` as this is already documented. But we could also have a more ergonomic subpath for type augmentation.

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
       "require": "./dist/module.cjs"
     },
     "./transformers": "./dist/runtime/transformers/index.js",
-    "./transformers/*": "./dist/runtime/transformers/*.js"
+    "./transformers/*": "./dist/runtime/transformers/*.js",
+    "./dist/runtime/types": "./dist/runtime/types/index.js"
   },
   "main": "./dist/module.cjs",
   "module": "./dist/module.mjs",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -16,8 +16,9 @@
       "import": "./dist/module.mjs",
       "require": "./dist/module.cjs"
     },
-    "./transformers": "./dist/runtime/transformers/index.mjs",
-    "./transformers/*": "./dist/runtime/transformers/*.mjs"
+    "./transformers": "./dist/runtime/transformers/index.js",
+    "./transformers/*": "./dist/runtime/transformers/*.js",
+    "./dist/runtime/types": "./dist/runtime/types/index.js"
```
danielroe, line 21: We could also add something like this, and document instead importing/augmenting `@nuxt/content/types`:

```suggestion
    "./types": "./dist/runtime/types/index.js",
    "./dist/runtime/types": "./dist/runtime/types/index.js"
```

farnabaz (approved):

PR #2495, opened 2024-01-09
chore: move to `shikiji` for types

antfu:
`@nuxt/mdc` has moved to `shikiji` and there is no use for `shiki-es` other than the type.


pi0:
shiki-es npm package is also now deprecated

farnabaz (approved):

Issue #2512, opened 2024-01-18
module loads both `js` and `mjs` files in client


farnabaz:
As of https://github.com/nuxt/content/commit/dff252b275228a3d9bef0309db6b741e07a6c7b9 the module loads both `mjs` and `js` files in client. This cause an issue in loading scoped data in files and create duplicate data.
<img width="197" alt="Screenshot 2024-01-18 at 15 35 02" src="https://github.com/nuxt/content/assets/2047945/cda0b97d-1938-4dee-9747-a1036c707fc0">

/cc @danielroe what do you think?

farnabaz:
The reproduction is tricky, but you can see in this link that there is two different `client-db` https://farnabaz-alpine-18.nuxt.space/articles/new/?preview=d4bd4346217e7c8c6d5c253d921113f2

Reproducing it with nuxt studio in development mode is easy, but other than that is tricky.
I've also checked with Content@2.9.0 and it works fine

PR #2514, opened 2024-01-19
build: do not emit `.mjs` in runtime directory

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/content/issues/2512

### ❓ Type of change

### 📚 Description

We probably need to fix this in nuxt/module-builder, but adding the new entrypoint didn't overwrite existing one, so both `.js` and `.mjs` files were being created.

### 📝 Checklist

### DIFF
--- a/build.config.ts
+++ b/build.config.ts
@@ -1,11 +1,18 @@
 import { defineBuildConfig } from 'unbuild'
 
 export default defineBuildConfig({
-  entries: [
-    {
-      input: 'src/runtime/',
-      outDir: 'dist/runtime',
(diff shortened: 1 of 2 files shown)


farnabaz:
Thanks @danielroe
What about module file? Why not removing `mjs` and `cjs` and just create `.js` extension?

farnabaz (approved):

danielroe:
I'd be very happy with that as well, but it shouldn't be linked to the bug and I wanted to keep the surface area of the change limited.

PR #2546, opened 2024-02-07
fix(types): use const type parameter for QueryBuilder.only

### LINKED ISSUE #2545: use const type Parameter for QueryBuilder.only() to not lose typing

When you're building a query and you use the `.only()` method you lose the typing because typescript only infers `Pick<ParsedContent, string>` and not `Pick<ParsedContent, "whatever-you-specified">`.
I had this problem when using queryContent("...").only(["...","..."]).findSurround()

The solution would be using `<const K extends keyof T>` instead of `<K extends keyof T>`
The const type parameter was introduced with Typescript 5.0

Letting it stay this way

thunfisch987:
### 🔗 Linked issue
#2545 
### ❓ Type of change

### 📚 Description

Resolves #2545 
This PR enables having better types when using the .only() function and uses the const type parameter feature that was introduced in Typescript 5.0

### 📝 Checklist

### DIFF
--- a/src/runtime/types/query.ts
+++ b/src/runtime/types/query.ts
@@ -280,8 +280,8 @@ export interface ContentQueryBuilder<T = ParsedContentMeta, Y = {}> {
   /**
    * Select a subset of fields
    */
-  only<K extends keyof T | string>(keys: K): ContentQueryBuilder<Pick<T, K>, Y>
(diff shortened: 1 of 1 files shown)


farnabaz (approved):
Thanks

thunfisch987:
I noticed this could probably also be done for .without(), but i didn't try it yet

PR #2673, opened 2024-06-20
fix(prerender): ensure `/` exists inside prerender rotues

farnabaz:
### ❓ Type of change

### 📝 Checklist


stefanprobst:
@farnabaz i think this breaks nuxt/i18n, which intentionally removes "/" from prerendering so redirecting to "/en" still works. see https://github.com/nuxt-modules/i18n/pull/2894

ptdev:
> @farnabaz i think this breaks nuxt/i18n, which intentionally removes "/" from prerendering

Confirming that it did indeed break i18n. See #2678

ydnar:
This PR, with its unilateral addition of `/` to the prerender list, broke production for us.

PR #2675, opened 2024-06-21
perf(dev-cache): improve localhost markdown page navigation performance (when having 2,000+ pages)

### LINKED ISSUE #2674: Markdown page navigation slow on dev server when having multiple thousand md pages

When having a few thousand markdown pages (2,000+), navigating between markdown pages when running the dev server can take seconds to load the new page.
(I observed multiple seconds on an Intel MacBook Pro and 1~2 seconds on a M1 Pro MacBook Pro per page transition.)

This problem explicitly occurs on the **local dev server**, for **markdown pages** only, and only when having **2,000+ markdown pages**, making local development quite frustrating.
Build times are fine though and navigating on the SSR generated site is snappy.

The problem seems to be caused by not using the content cache on localhost.
https://github.com/nuxt/content/blob/3a2b5d1ec19f9232dd9f5b3487c93ff3997c7476/src/runtime/server/storage.ts#L147-L149

I assume not using the cache keeps hot reload working when files are changed, however this seems to cause the severe degradation in performance when dealing with lots of markdown pages.

To fix the issue I suggest using the cache on localhost as well, and flushing the cache when a file change is detected, to keep hot reload working.
I will create a PR (draft) later to address this.

N/A

### Additional context

Video showing the issue:

https://github.com/nuxt/content/assets/7602082/7582f494-1fd4-43b2-aaac-8ebc1dd8c7a6

On the video loading times are still < 1s, however the situation gets worse for more complex markdown pages.

Video showing the fix I'm working on:

https://github.com/nuxt/content/assets/7602082/15f27e2a-6923-4cfd-8581-62e831065fb8

## Reproduction

I've created a project on StackBlitz to reproduce the issue.
https://stackblitz.com/~/github.com/coffeephile/nuxt-huge-project-test

1. Make sure to check out branch `nuxt-content-performance-patch`
2. Install packages `npm i`
3. Run `NUXT_NUM_PAGES=10000 npm run dev` (generate 10,000 markdown pages)
4. Open preview in new tab (or better: run locally on your machine)

To apply the fix:
1. Run `chmod +x patch-content-localhost.sh`
2. Run `./patch-content-localhost.sh`
3. Run `NUXT_NUM_PAGES=10000 npm run dev`
4. Open preview in new tab (or better: run locally on your machine)

coffeephile:
### 🔗 Linked issue

https://github.com/nuxt/content/issues/2674

### ❓ Type of change

### 📚 Description

Resolves #2674 

Navigating between markdown pages on localhost (dev server) is very slow when dealing with 2,000+ markdown pages.
(Please refer to linked issue for demo and reproduction)

By using `cachedContents` on localhost as well (not only during build time) this issue can be resolved.
In order to keep hot reload working, the cache gets flushed when a file change is detected.

I confirmed my fix to be working for my use case, however I'm not sure if I am breaking anything else in the process, so any advice or help is welcome.
This PR should be considered a first draft, rather than a perfect solution. 🙇‍♂️ 

### 📝 Checklist


src/module.ts
```diff
@@ -447,6 +448,7 @@ export default defineNuxtModule<ModuleOptions>({
           ? './plugins/documentDriven'
           : './legacy/plugins/documentDriven'
       ))
+      addServerPlugin(resolveRuntimeModule("./server/plugins/refresh-cache"));
```
farnabaz, line 451: We should add this plugin only in development mode

src/runtime/server/storage.ts
```diff
@@ -144,16 +147,14 @@ export const getContentsList = (() => {
     if (event.context.__contentList) {
       return event.context.__contentList
     }
-    if (isPrerendering && cachedContents.length) {
+    if (cachedContents.length) {
```
farnabaz, line 150: Shall we keep `isPrerendering` and add a condition for dev?

```suggestion
    if ((isPrerendering || !isProduction) && cachedContents.length) {
```

src/runtime/server/storage.ts
```diff
@@ -144,16 +147,14 @@ export const getContentsList = (() => {
...
       return cachedContents
     }
 
     if (!pendingContentsListPromise) {
       pendingContentsListPromise = _getContentsList(event, prefix)
       pendingContentsListPromise.then((result) => {
-        if (isPrerendering) {
-          cachedContents = result as ParsedContent[]
-        }
+        cachedContents = result as ParsedContent[]
```
farnabaz, line 157: Here we can add back prerendering and check for dev

farnabaz (changes requested):
By removing the prerendering check, we are changing the behavior of the production environment too.
We should add back prerendering check and add logic to handle dev mode

coffeephile:
@farnabaz Many thanks for your swift review, much appreciated! 🙏 

I've implemented the changes you suggested, please check again 🙇‍♂️

farnabaz (approved):
LGTM 👍 
Thanks

coffeephile:
Thank you! 🙏😊

Issue #437, opened 2020-08-31
How do you return posts created in a date range?


davydnorris:
Hi all,

I am trying to implement an archives search in my blog, which requires me to return posts within a date range. I am trying the following but with no luck:

```
        let postQuery = this.$content('blog', { deep: true })
          .only(['dir', 'title', 'author', 'description', 'image', 'createdAt'])
          .sortBy('createdAt', 'desc');

        if (this.dateFilter) {
          const startDate = new Date(this.dateFilter.setDate(1));
          const endDate = new Date(this.dateFilter.setMonth(this.dateFilter.getMonth() + 1, 1));
          postQuery = postQuery.where({ createdAt: { $between: [startDate, endDate] } });
        }

        this.posts = await postQuery.fetch();
```
the above gives no errors, but just doesn't return anything. I have tried all sorts of variations on this, such as:
```
          postQuery = postQuery.where({
            $and: [
              { createdAt: { $gte: startDate } },
              { createdAt: { $lte: endDate } }
            ]
          });
```
but still no luck. The only thing I can't think of is that createdAt is not a date?

Any help much appreciated!

azrikahar:
This is not the final solution, but I'm facing some weird interaction regarding the date here.  It works upon refresh on `blog` page, but using button to trigger fetch will fail instead. CodeSandbox link: https://codesandbox.io/s/happy-jepsen-1x3mo?file=/pages/blog/index.vue

## Scenario

blog posts:
- Post 1 is created at 2020-08-25
- Post 2 is created at 2020-08-31

I'm using the Nuxt fetch() method instead of asyncData() here so we can trigger $fetch on button click. If we were to go to `/blog` via links or routes, we will see the fetch returning empty, meaning the filtering is incorrect. The fetch didn't fail because it would show `Unable to fetch blog posts` if it did.

However, if you go to `/blog`, then do a refresh by clicking on the refresh button on the preview box (not our browser, the sandbox preview), you can see the fetch worked and it returned only Post 2 because I hardcoded the initial start date as `2020-08-26` which is greater than the Post 1's `2020-08-25`.

I've also noticed if the fetch happened on the server, it just needs `new Date(date)` to console.log an ISO date format, whereas the client side fetch will need `new Date(date).toISOString()` to have the same format. I thought having the same format for client side will fix it, but it still fails :/ Not really sure what is happening here. 

[...]

PR #2704, opened 2024-07-11
fix: add string type to gt/lt/gte/lte type definitions

damieng:
### 🔗 Linked issue

https://github.com/nuxt/content/issues/437

### ❓ Type of change

### 📚 Description

The `$gte`, `$gt`, `$lt`, `$lte` operators are defined in the Typescript as only taking a number. This is in fact incorrect as they map to the standard JavaScript `>=`, `>`, `<`, `<=` operators which can handle any more types and as the test coverage shows they work perfect with strings.

This fix is needed as anyone trying to use these operators with a string - such as in a common case as filtering by alpha ranges or even a sort-formatted date string - will receive Typescript errors leading them to think this is not the way to do it.

This adds the `string` type to those operators, adds test coverage to prove they work and importantly adds an example to the filtering docs showing how to use the string filtering to sort by a date which is a common situation and one the linked ticket went into.

### 📝 Checklist

### DIFF
--- a/docs/content/7.v1/1.getting-started/4.fetching.md
+++ b/docs/content/7.v1/1.getting-started/4.fetching.md
@@ -76,6 +76,8 @@ const articles = await this.$content('articles').where({ title: { $eq: 'Home' } 
 
 // $gt
 const articles = await this.$content('articles').where({ age: { $gt: 18 } }).fetch()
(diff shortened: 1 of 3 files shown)


farnabaz (approved):
Thanks

PR #2891, opened 2024-12-06
feat(config): provide defineContentConfig utility

ineshbose:
### ❓ Type of change

### 📚 Description

This is NOT a breaking change.

Devs can still use `export const collection` (as `c12` will fallback non-default export). However, in consistency with other `.config` files that tend to use default export and providing type-safety, having `defineContentConfig` would be nice.

### 📝 Checklist

### DIFF
--- a/docs/content.config.ts
+++ b/docs/content.config.ts
@@ -1,24 +1,26 @@
-import { defineCollection, z } from '@nuxt/content'
+import { defineContentConfig, defineCollection, z } from '@nuxt/content'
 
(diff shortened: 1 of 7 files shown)


ineshbose:
Note: I've just added ability to use `defineContentConfig` without needing to import, but it may be a bit experimental. Let me know if you want to exclude from this PR for now.

farnabaz (approved):
LGTM 👍 

Thanks

Issue #2921, opened 2024-12-18
Add option for Content Collections to access raw file content


maximepvrt:
## Description
In the current alpha version of `@nuxt/content` v3, it seems that hooks have been removed, and accessing the raw content of files has become impossible without them (which, to be fair, didn’t work properly in v2 either). For certain use cases, it is crucial to retrieve the raw file content before parsing it with `parseMarkdown()` or other custom transformations.

This feature has been a recurring request in the community, as evidenced by issues like #2803 and #2490. Adding a direct method or option to retrieve the raw file content would enhance the flexibility of the module and cater to advanced use cases.

## Use Case
In my specific case, I need to process the raw file content in my page before parsing it with `parseMarkdown()`.

oneminch:
According to #2925, these hooks are now available in `v3`. They are available in build time (instead of runtime using server plugins as in `v2`). Here is the documentation: https://content3.nuxt.dev/docs/advanced/hooks. @maximepvrt

maximepvrt:
Thank you for the update and the link to the documentation! It's great to see the return of hooks in v3 🎉.

However, I’ve noticed that these hooks don’t work with Nuxt Studio, even if the project is not configured as a static build. The issue seems to be that the Content API is replaced by Studio's API, which prevents hooks from functioning properly.

This is precisely why I’m suggesting the addition of a raw option in the Content Collections API. Having this option would provide a straightforward way to access raw file content, regardless of the environment (Studio or not), and would address scenarios where hooks are not a viable solution.

oneminch:
That's similar to what I’m suggesting in #2803.

farnabaz:
In version 3 you can define `rawbody` field in the collection's schema to store raw content in the database. checkout https://content3.nuxt.dev/docs/advanced/raw-content
The integration with the studio for `rawbody` is still in progress.

oneminch:
Does that mean raw content can also be accessed from an `afterParse` hook if the field is defined in the collection schema?

farnabaz:
Yes, the module will fill `rawbody` before calling the hook.

maximepvrt:
So cool ! Thanks @farnabaz @larbish 💚

PR #2925, opened 2024-12-20
feat: pre / post parsing hooks

harlan-zw:
### 🔗 Linked issue

https://github.com/nuxt/content/issues/2921

### ❓ Type of change

### 📚 Description

The runtime hooks from Content v2 were useful for modules to provide an integration with markdown files. For example, allowing users to provide custom frontmater keys or inferring data from the markdown body.

- [Nuxt Sitemap - Nuxt Content Integration](https://github.com/nuxt-modules/sitemap/blob/main/src/runtime/server/plugins/nuxt-content.ts)

In Content v3 these runtime hooks are moved to the build time which is a great performance win, however, we no longer have access to the parsing hooks. 

This PR introduces the `collection:parsedFile` hook to allow modules to start integrating with v3. 

### 📝 Checklist

### DIFF
--- a/docs/content/docs/7.advanced/5.hooks.md
+++ b/docs/content/docs/7.advanced/5.hooks.md
@@ -45,7 +45,7 @@ export interface FileAfterParseHook {
 ```ts
 export default defineNuxtConfig({
   hooks: {
-    'content:file:afterParse'(ctx) {
+    'content:file:afterParse'(ctx: FileAfterParseHook) {
       // ...
     }
   }
--- a/docs/content/docs/7.advanced/5.hooks.md
+++ b/docs/content/docs/7.advanced/5.hooks.md
@@ -5,14 +5,14 @@ navigation:
   title: Hooks
 ---
 
-## `content:parsed`{lang="ts"}
+## `collection:parsedFile`{lang="ts"}
 
 This hook is called after the content is parsed and before it is saved to the database.
 
 ```ts
 export default defineNuxtConfig({
   hooks: {
-    'content:parsed'(ctx) {
+    'collection:parsedFile'(ctx) {
       // { content: CollectionItemBase, collection: ResolvedCollection }
     }
   }
--- a/docs/content/docs/7.advanced/5.hooks.md
+++ b/docs/content/docs/7.advanced/5.hooks.md
@@ -13,14 +13,8 @@ It can be used to modify the raw content from a `file` before it is transformed
 or modify the transform options.
 
 ```ts
-export interface FileBeforeParseHook {
-  collection: ResolvedCollection
-  file: ContentFile
-  parserOptions: TransformContentOptions
-}
-```
+import type { FileBeforeParseHook } from '@nuxt/content'
 
-```ts
 export default defineNuxtConfig({
   hooks: {
     'content:file:beforeParse'(ctx: FileBeforeParseHook) {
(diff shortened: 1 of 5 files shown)


farnabaz:
Thanks for the PR

Can we also call a hook before parsing the file with raw markdown? There is also a feature request for `beforeParse` hook in v3 https://github.com/nuxt/content/issues/2921

Also let's rename the hooks and add `content:`  prefix to emphasize the hook's owner and scope.
- `content:file:beforeParse` (I think using `:file:` instead of `:collection:` makes more sense)
- `content:file:afterParse`

It is worth mentioning that the `body` field in v3 is minified and the tree structure is different from v2 so we might need to explain the new minified structure. (we can add this in a separate PR)

harlan-zw:
Thanks for the quick review! I originally tried to use different hook names so that the breaking changes but I prefer your suggestions.

It would be nice to have the content from afterParse typed but I don't think there's a way to do that unless we introduce collection based hooks but the complexity probably isn't worth pursuing at this stage.

Let me know if you'd like anything else updated.

src/utils/content/index.ts
```diff
@@ -131,13 +133,17 @@ export async function createParser(collection: ResolvedCollection, nuxt?: Nuxt)
...
+    const { file: hookedFile, collection: hookedCollection } = beforeParseCtx
 
-    const parsedContent = await transformContent(file, parserOptions)
+    const parsedContent = await transformContent(hookedFile, beforeParseCtx.parserOptions)
     const { id: id, ...parsedContentFields } = parsedContent
     const result = { id } as typeof collection.extendedSchema._type
     const meta = {} as Record<string, unknown>
 
-    const collectionKeys = Object.keys(collection.extendedSchema.shape)
+    const collectionKeys = Object.keys(hookedCollection.extendedSchema.shape)
```
farnabaz, line 146:
```suggestion
    const collectionKeys = Object.keys(collection.extendedSchema.shape)
```

it's more secure to use `collection` rather than `hookedCollection`. Collections in these hooks should be readonly

docs/content/docs/7.advanced/5.hooks.md
```diff
@@ -0,0 +1,53 @@
...
+  collection: ResolvedCollection
+  file: ContentFile
+  content: ParsedContentFile
+}
+```
+
+```ts
+export default defineNuxtConfig({
+  hooks: {
+    'content:file:afterParse'(ctx) {
```
farnabaz, line 48:
```suggestion
    'content:file:afterParse'(ctx: FileAfterParseHook) {
```

docs/content/docs/7.advanced/5.hooks.md
```diff
@@ -0,0 +1,53 @@
...
+  collection: ResolvedCollection
+  file: ContentFile
+  parserOptions: TransformContentOptions
+}
+```
+
+```ts
+export default defineNuxtConfig({
+  hooks: {
+    'content:file:beforeParse'(ctx) {
```
farnabaz, line 26:
```suggestion
    'content:file:beforeParse'(ctx: FileBeforeParseHook) {
```

farnabaz (commented):
Looks good to me 👍 just suggested small changes

src/utils/content/index.ts
harlan-zw: I wasn't exactly sure how all of my modules integrating with Nuxt Content v3 would work so I wanted to be able to modify a collection schema using a hook if needed.

Thinking on your comment, I now believe this would be better done via a dedicated hook if it is needed. I can send a PR if / when that blocks me.

farnabaz (approved):

PR #2829, opened 2024-10-30
chore(config): use c12

ineshbose:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Use c12 to load config - it would give lot of other benefits. Let me know if using custom loader was intentional! 🙂 

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #2929, opened 2024-12-24
feat(config): enable watch for all layer configs and load in parallel

ineshbose:
### ❓ Type of change

### 📚 Description

Another small PR inspired from my reading of the implementation for v3..

A few bits I had noticed:

After #2829, we would have enabled `.config/` directory (or even config in different extensions) - but I realised that the fsStat before loading the config would not allow us for this. So it started with that, then I went ahead to parallelise the call to `loadContentConfig`, and then resolve `collectionMap`. I realised that calling `resolveCollections` and iterating over collections was not necessary (as the key is the `collection.name`), so instead we use a reducer to override any low-level layer collections (ideally we would have wanted to use `defu`, but that is deep-default and may have side-effects for collection schemas). In addition to that, I also ensured that the configurations would work with a watcher, so thanks for `watchConfig`, all of this is possible.

Let me know what you think - I had to spend some time scratching my head on updating this code, so would love feedback! 🙂 

### 📝 Checklist

### DIFF
--- a/src/utils/config.ts
+++ b/src/utils/config.ts
@@ -50,7 +50,7 @@ export async function loadContentConfig(nuxt: Nuxt, options: { defaultFallback?:
   const collectionsConfig = contentConfigs.reduce((acc, curr) => ({ ...acc, ...curr.config?.collections }), {} as Record<string, DefinedCollection>)
   const hasNoCollections = Object.keys(collectionsConfig || {}).length === 0
 
-  if (!hasNoCollections) {
+  if (hasNoCollections) {
     logger.warn('No content configuration found, falling back to default collection. In order to have full control over your collections, create the config file in project root. See: https://content.nuxt.com/getting-started/installation')
   }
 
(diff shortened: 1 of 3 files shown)


src/utils/config.ts
```diff
@@ -21,47 +20,41 @@ const defaultConfig: NuxtContentConfig = {
...
-  for (const layer of _layers) {
-    const rootDir = layer.config.rootDir
-    const configStat = await stat(join(rootDir, 'content.config.ts')).catch(() => null)
-    if (configStat && configStat.isFile()) {
-      const { collections } = await loadContentConfig(rootDir, { defaultFallback: false })
-      for (const collection of collections) {
-        collectionMap[collection.name] = collection
-      }
-    }
+  if (!hasNoCollections) {
```
farnabaz, line 53:
```suggestion
  if (hasNoCollections) {
```

farnabaz (approved):
LGTM 👍 
Thanks

PR #2934, opened 2024-12-30
fix: add module entry points for TypeScript and module imports

productdevbook:
### ❓ Type of change

### 📚 Description

With Nuxt 3.15 it seems that if there is a module auto-discovery feature, it seems to be turned off. I added this and then the module started working.

### 📝 Checklist


farnabaz:
Please provide a reproduction sample for the issue you were facing.
Module should work fine without this export

productdevbook:
I can't reproduce reproduction for some reason, it seems to be a computer thing. It's only solved if I do this.

PR #2985, opened 2025-01-16
docs(PostPage): fix surround order

maximepvrt:
### ❓ Type of change

### 📚 Description

@farnabaz There is still an issue with the surround links being in the wrong order at the bottom.

Capture d’écran 2025-01-16 à 18 09 40

### 📝 Checklist

### DIFF
--- a/docs/app/components/post/PostPage.vue
+++ b/docs/app/components/post/PostPage.vue
@@ -16,7 +16,7 @@ const { data } = await useAsyncData(route.path, () => Promise.all([
   queryCollectionItemSurroundings('posts', route.path, { fields: ['title', 'description'] })
     .where('path', 'LIKE', `/${type}%`)
     .where('draft', '=', 0)
-    .order('date', 'ASC'),
+    .order('date', 'DESC'),
 ]), {
   transform: ([page, surround]) => ({ page, surround }),
 })


farnabaz:
Isn't this true order? I think previous items should be on the left side. They are similar to the back button just above them.

Correct me if I'm wrong @benjamincanac

maximepvrt:
@farnabaz All previous links are currently on the right, and next links are on the left. You can see an example of this behavior on this page: https://nuxt.com/blog/v3-15. It would work perfectly if the order followed the same logic as listing all the articles here: https://github.com/nuxt/content/blob/main/docs/app/pages/blog/index.vue#L14.

However, it might make sense to reverse the order if the locale uses a right-to-left (RTL) reading direction. What do you think?

farnabaz:
It is worth waiting for @benjamincanac opinion about this. As I see in `ContentSurround` component, the previous item should have `direction="left"` attribute which matches the current order.


```vue
    <ReuseLinkTemplate :link="surround[0]"... direction="left" />
    <ReuseLinkTemplate :link="surround[1]" ... direction="right" />
```

benjamincanac (approved):

farnabaz:
Well, I had a quick chat with Benjamin, and seems that I got it wrong. 
Thanks for the PR @maximepvrt ❤️

PR #3018, opened 2025-01-22
fix: convert boolean value to number in query condition

### LINKED ISSUE #2927: Zod not working correcly in V3 schema
### Environment

Nuxt project info:                                                                                                                                                                         11:18:23 AM

------------------------------
- Operating System: Linux
- Node Version:     v20.12.2
- Nuxt Version:     3.14.1592
- CLI Version:      3.17.1
- Nitro Version:    2.10.4
- Package Manager:  yarn@1.22.22
- Builder:          -
- User Config:      default
- Runtime Modules:  @nuxt/ui-pro@3.0.0-alpha.10, @nuxtjs/i18n@9.1.1, @nuxt/content@3.0.0-alpha.8
- Build Modules:    -
------------------------------

### Reproduction

1. Using the config below: content.config.ts
```typescript
import {defineCollection, defineContentConfig, z} from '@nuxt/content'

export default defineContentConfig({
    collections: {
        blog: defineCollection({
            // Load every file inside the `content` directory
            source: 'blog/**',
            // Specify the type of content in this collection
            type: 'page',
            schema: z.object({
                badge: z.string(),
                image: z.string(),
                tags: z.array(z.string()),
                published: z.boolean(),
                createdAt: z.string().datetime(),
                updatedAt: z.string().datetime(),
            })
        }),
    }
})
```
 
2. Code
```typescript
        if (options.published !== undefined) {
            // zod use 1 and 0 as boolean
            const published = options.published ? 1 : 0;
            query = query.where("published", '=', published);
        }
        if (options.latest !== undefined) {
            query = query.order('updatedAt', "DESC")
        }
```
Image

3. Markdown
```markdown
---
title: Example title
description: This is example description
badge:
image: /assets/images/coming-soon.jpg
tags:
  - Example
published: false
createdAt: 2025-01-01 00:00
updatedAt: 2025-01-01 00:00
---

## Example title

Nothing here yet. Comeback later.
```

**Bugs:**
**1. order() not able to use the field in `schemas`**
**2. z.boolean() result to numberr `0/1` not boolean `false/ true`**

### Describe the bug

**1. Bug 1:**  order() not able to use the field in `schemas`
When using order with field in schemas it show the below error
> Argument of type 'string' is not assignable to parameter of type 'keyof Collections[T]'.
  Type '"updatedAt"' is not assignable to type '"extension" | "path" | "title" | "description" | "seo" | "body" | "navigation" | "id" | "stem" | "meta"'.
Image

**2. Bug 2:** z.boolean() result to numberr `0/1` not boolean `false/ true`
Image

### Additional context

I also see that nuxt/content don't have any `total` attribute when get from collection so that I can't create paginate and skip() will be useless because it can't count and paginate

### Logs

```Shell

```

HADB:
### 🔗 Linked issue
https://github.com/nuxt/content/issues/2927

### ❓ Type of change

### 📚 Description

After some investigation, I discovered that #2927 is caused by the following code in `query.ts`, where the `value` is wrapped in quotes. This prevents boolean results from being found. 

https://github.com/nuxt/content/blob/d5efff1fdd03612d712661bae7f4596965082e44/src/runtime/internal/query.ts#L52

In SQLite, boolean values are stored as `0` and `1`. When searching with `true` or `false`, they are converted to `0` and `1`, which works fine. However, when searching with `'true'` or `'false'`, it fails.

The solution is to convert them to `0` and `1` before adding quotes.

This should work with both `0/1` and `true/false` in query condition.

Resolves #2927

### 📝 Checklist

### DIFF
--- a/docs/content/docs/4.utils/2.query-collection-navigation.md
+++ b/docs/content/docs/4.utils/2.query-collection-navigation.md
@@ -30,7 +30,7 @@ The function returns a chainable promise that allows you to add additional query
 <script setup lang="ts">
 const { data } = await useAsyncData('navigation', () => {
   return queryCollectionNavigation('docs')
-    .where('published', '==', true)
+    .where('published', '=', true)
     .order('date', 'DESC')
 })
 </script>
@@ -81,8 +81,8 @@ Example with additional query conditions and extra fields:
 <script setup lang="ts">
 const { data } = await useAsyncData('navigation', () => {
   return queryCollectionNavigation('docs', ['description', 'badge'])
-    .where('draft', '==', false)
-    .where('partial', '==', false)
+    .where('draft', '=', false)
+    .where('partial', '=', false)
     .order('title', 'ASC')
 })
 </script>
--- a/docs/content/docs/4.utils/3.query-collection-item-surroundings.md
+++ b/docs/content/docs/4.utils/3.query-collection-item-surroundings.md
@@ -31,7 +31,7 @@ The function returns a chainable promise that allows you to add additional query
 <script setup lang="ts">
 const { data } = await useAsyncData('surround', () => {
   return queryCollectionItemSurroundings('docs', '/foo')
-    .where('published', '==', true)
+    .where('published', '=', true)
     .order('date', 'DESC')
 })
 </script>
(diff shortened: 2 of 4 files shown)


farnabaz (approved):
Thanks 👍

PR #3121, opened 2025-02-11
fix: `queryCollectionItemSurroundings` type definition in built module

### LINKED ISSUE #3120: `queryCollectionItemSurroundings` type definition broken in `app.d.ts`
### Environment

- Operating System: `Darwin`
- Node Version:     `v20.17.0`
- Nuxt Version:     `3.15.2`
- CLI Version:      `3.20.0`
- Nitro Version:    `2.10.4`
- Package Manager:  `pnpm@9.9.0`
- Builder:          `-`
- User Config:      `compatibilityDate`, `devtools`, `typescript`, `vite`, `alias`, `devServer`, `experimental`, `nitro`, `modules`, `fonts`, `build`, `imports`, `css`, `runtimeConfig`, `content`
- Runtime Modules:  `@nuxthq/studio@2.2.1`, `@nuxt/content@3.0.0`, `@nuxt/icon@1.10.3`, `@vueuse/nuxt@12.5.0`, `@nuxt/fonts@0.10.3`, `@nuxt/test-utils/module@3.15.4`, `@nuxt/image@1.9.0`, `@nuxt/scripts@0.9.5`
- Build Modules:    `-`

### Version

v3

### Reproduction

Image

### Description

There appears to be a bug with the type extraction being used by `nuxi` when extracting the type definition for the `queryCollectionItemSurroundings`.

It does not include the import for `ContentNavigationItem` for some reason. This means that the returned type from `queryCollectionItemSurroundings` is broken and appears to have no properties on it.

### Additional context

_No response_

### Logs

```shell-script

```

GerryWilko:
### 🔗 Linked issue

fix #3120

### ❓ Type of change

### 📚 Description

This resolves the broken type definition for `queryCollectionItemSurroundings` in the built module. I am unsure why `nuxi` is building a broken TS file but adding an explicit import and then export (to stop it being tree-shaken) seems to fix this issue with the `ContentNavigationItem` being incorrectly bundled.

### 📝 Checklist

### DIFF
--- a/src/runtime/app.ts
+++ b/src/runtime/app.ts
@@ -1,4 +1,4 @@
-import type { Collections, PageCollections, CollectionQueryBuilder, SurroundOptions, SQLOperator, QueryGroupFunction } from '@nuxt/content'
+import type { Collections, PageCollections, CollectionQueryBuilder, SurroundOptions, SQLOperator, QueryGroupFunction, ContentNavigationItem } from '@nuxt/content'
 import type { H3Event } from 'h3'
 import { collectionQueryBuilder } from './internal/query'
 import { generateNavigationTree } from './internal/navigation'
--- a/src/runtime/app.ts
+++ b/src/runtime/app.ts
@@ -7,6 +7,10 @@ import { generateSearchSections } from './internal/search'
 import { fetchQuery } from './internal/api'
 import { tryUseNuxtApp } from '#imports'
 
+export type {
+  ContentNavigationItem,
+}
+
 interface ChainablePromise<T extends keyof PageCollections, R> extends Promise<R> {
   where(field: keyof PageCollections[T] | string, operator: SQLOperator, value?: unknown): ChainablePromise<T, R>
   andWhere(groupFactory: QueryGroupFunction<PageCollections[T]>): ChainablePromise<T, R>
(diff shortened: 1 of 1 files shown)


src/runtime/app.ts
```diff
@@ -7,6 +7,10 @@ import { generateSearchSections } from './internal/search'
 import { fetchQuery } from './internal/api'
 import { tryUseNuxtApp } from '#imports'
 
+export type {
+  ContentNavigationItem,
+}
+
```
farnabaz, line 13: Can't update `queryCollectionItemSurroundings` and explicitly point to type instead of exporting type here?

```ts

import type { ContentNavigationItem } from '@nuxt/content'
export function queryCollectionItemSurroundings<T extends keyof PageCollections>(collection: T, path: string, opts?: SurroundOptions<keyof PageCollections[T]>): ChainablePromise<T, ContentNavigationItem[]> {
  return chainablePromise(collection, qb => generateItemSurround(qb, path, opts))
}
```

I guess this resolves the issue without exporting the type here. The idea is to keep all type exports in same place

GerryWilko: I've updated the return types of those two functions instead.

farnabaz (approved):
Thanks 👍

PR #3144, opened 2025-02-17
test: track down origin of queries (do not merge yet)

elevatebart:
This PR is to track down where concurent initializations come from


elevatebart:
OK it's the ready flag that comes out as a number instead of a boolean... life sucks

PR #3145, opened 2025-02-17
fix: rollback changes to syntax

elevatebart:
kestra.io is down right now and we need a fix quick
This is my first attempt

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -246,9 +246,10 @@ async function processCollectionItems(nuxt: Nuxt, collections: ResolvedCollectio
     collectionDump[collection.name] = []
 
     // we start by telling everyone that we are setting up the collection
+    // so we create a new entry in the info table saying that it is not ready yet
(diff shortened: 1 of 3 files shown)


atinux:
Cannot you lock the version of `@nuxt/content` in the meantime for the website to be up?

elevatebart:
The official version of content 3 breaks the site (see concurrent initialization PR).
To have a working website and avoid the rework of the website backwards, I made a quick change. 
The fix worked for the whole week end.

@farnabaz had a few remarks that I applied and when using the updated code nothing works anymore:
- there seems to be two versions of the content fighting for initialization on the prod database
- Could this be cache related?
- Is it due to the chnage in how the hash is calculated?
Anyway, I ma trying to figure out this concurency issue as fast as I can.

Thanks for all the help you can bring.

elevatebart:
OK it's the ready flag that comes out as a number instead of a boolean... life sucks

PR #3155, opened 2025-02-18
fix(database): check that table exist before querying

elevatebart:
The idea here is to avoid catching errors so that if there is an error, it's not hidden.
The try/catch is troublesome here and makes me worried.

To that end I make sure that the checks are done manually before starting the queries.

Also I figured out why the info ready was not converted: ` FROM _content_info` has to be all caps for it to match.
Fixed this as well.


elevatebart:
OK this PR is moot since you added the `databaseVersion`. 
Closing.

I would still prefer that we check the existence of the info table before select so that, in case of a real error, the system logs it.

PR #3158, opened 2025-02-19
feat: page level caching

### LINKED ISSUE #3151: Caching on page level

When website maintainers update anything in the blog, the blog takes forever to load.
That is because all data in the base is `DROPPED` even if most of the data has not changed. 

Only `DROP` the table if the structure has changed.
Instead of `DROP`ping the existing table, we should delete the changed records and add those missing.

To that end: 

- **Build time** 
  - Only take into account the definition of the table in the table hash
  - Add a `__hash__` field on each content table to store the hash of the current record
    - it should have the "unique" constraint
  - Calculate at build time the hash of the values then append the hash at the end of the values before setting the query
  - Store the list of hash in the database dump as the first line of the dump. 
    - Store in a SQL comment to respect sql syntax
    - Store it as a JSON table to facilitate parsing and avoid bad chars
- **At runtime**
  - Start by getting the list of existing hash from the database
  - From the database dump get the list of hash that needs to be set after init
  - Get the list of hash that need to be removed and delete them (split into multiple queries if necessary)
  - Run the needed inserts from the dump
    - Since the queries and the hash are in the same order, use the array to know what inserts to run

We could use smarter caching groups for collections, like grouping cache by date or by category. 
But that is more complicated to setup and less absolute than the above.

We can also use an update on id since it is the primary key for the relevant records. But, from experience, updates are slower than inserts.

### Additional context

none

elevatebart:
### 🔗 Linked issue

closes #3151

### ❓ Type of change

### 📚 Description

This PR moves the caching level to each page/record instead of each collection.

This allows for faster iteration.

- **Build time** 
  - Only take into account the definition of the table in the table hash
  - Add a `__hash__` field on each content table to store the hash of the current record
    - it should have the "unique" constraint
  - Calculate at build time the hash of the values then append the hash at the end of the values before setting the query
  - Store the list of hash in the database dump as the first line of the dump. 
    - Store in a SQL comment to respect sql syntax
    - Store it as a JSON table to facilitate parsing and avoid bad chars
- **At runtime**
  - Start by getting the list of existing hash from the database
  - From the database dump get the list of hash that needs to be set after init
  - Get the list of hash that need to be removed and delete them (split into multiple queries if necessary)
  - Run the needed inserts from the dump
    - Since the queries and the hash are in the same order, use the array to know what inserts to run

### 📝 Checklist


elevatebart:
This PR still needs some extra tests for updating an existing db.

elevatebart:
Live testing has been made on [this PR](https://github.com/kestra-io/docs/pull/2252).

I would love some help for E2E testing of updates.
I can't find a way to setup a project, build it is serve it... specially since I am trying to build/serve multiple times with file changes.

If you have the time to show me the ropes, I am all ears.
If not, I'd rather have you write the E2E tests here.

Test scenarii:
- initialization works on an empty database
- when collection structure change (add a new field), table is dropped
- when adding md doc, the new doc is added in the DB
- when removing a doc, removed in the DB
- when updating a doc, doc ...
- All 3 together (add some, remove some, update some)

To be complete, docs should be in multiple directories so re-order has an effect.

src/runtime/internal/database.server.ts
```diff
@@ -113,20 +115,64 @@ async function _checkAndImportDatabaseIntegrity(event: H3Event, collection: stri
...
+
+    // skip sql comment
+    if (sql.startsWith('-- ')) {
+      return Promise.resolve()
+    }
+
+    // If the structure has not changed,
+    // skip any insert/update line whose hash is already in the database.
+    // If not, since we dropped the table, no record is skipped, insert them all again.
+    if (unchangedStructure && (sql.startsWith('INSERT ') || sql.startsWith('UPDATE ')) && !indexesToInsert.includes(index)) {
```
farnabaz, line 165: Values of `indexesToInsert` does not match the index of queries in `dump` therefore `indexesToInsert.includes(index)`  is not a valid condition to check

farnabaz (changes requested):
Generally, I like the idea.
I have concerns about the solution. Removing drop queries is a breaking change for the full dump and breaks NuxtHub deployment. (NuxtHub uses this full dump to refill the database)

What if we keep the drop query and prefix it with structure checksum `/* checksum */ DROP ...`, then we ignore these lines if the checksum is not changed.

We can add similar prefixes for insert and update queries. This will simply hash detection here
```
if (unchangedStructure && (sql.startsWith('INSERT ') || sql.startsWith('UPDATE ')) && !indexesToInsert.includes(index)) {
```

The dump can be like

```sq;
/* structure:{HASH} */ DROP TABLE ......
/* structure:{HASH} */ CREATE TABLE ...
/* hash:{ROW_HASH} */ INSERT ...
/* hash:{ROW_HASH} */ UPDATE ...
```

src/runtime/internal/database.server.ts
elevatebart: The hashes indexed here come directly from the `dump`
1. So they are ordered the same way as the insert query, I make sure of it at build time.
2. The table of hash is padded at the beginning with the exact number of lines in the init. If there are 3 lines of insert, we insert 3 empty strings at the beginning.
3. That is why I would have loved a few e2e tests.
4. I like your idea of having a hash at the begining of the line so the code is simpler to understand

elevatebart:
I have tweaked the system a little:
- restored the drop table in the dump
- added the hash in comment in front of updates and inserts
- ignored all structure queries if the structure has not changed
- made sure that info version inserts and updates are always committed

elevatebart:
In that last commit I changed the format of the dump once more:

```sql
CREATE TABLE IF NOT EXISTS _content_info (id TEXT PRIMARY KEY, "ready" BOOLEAN, "structureVersion" VARCHAR, "version" VARCHAR, "__hash__" TEXT UNIQUE); /* structure */
INSERT INTO _content_info VALUES ('checksum_content', false, 'RYgZDcy6sk', 'v3.2.3--DZOppTJAd2', 'synekhxovA');
DROP TABLE IF EXISTS _content_content; /* structure */
CREATE TABLE IF NOT EXISTS _content_content (id TEXT PRIMARY KEY, "title" VARCHAR, "arrayField" TEXT, "body" TEXT, "booleanField" BOOLEAN, "description" VARCHAR, "extension" VARCHAR, "meta" TEXT, "navigation" TEXT DEFAULT true, "numberField" INT, "path" VARCHAR, "seo" TEXT DEFAULT '{}', "stem" VARCHAR, "__hash__" TEXT UNIQUE); /* structure */
INSERT INTO _content_content VALUES ('content/index.md', 'Home page', '["item1","item2"]', '{"type":"minimal","value":[["h1",{"id":"home-page"},"Home page"]],"toc":{"title":"","searchDepth":2,"depth":2,"links":[]}}', true, '', 'md', '{}', 'true', 1, '/', '{"title":"Home page","description":""}', 'index', 'JApZF0Dncz'); /* checksum: JApZF0Dncz */
UPDATE _content_info SET ready = true WHERE id = 'checksum_content'; 
```

This feels more readable to bothe human beings and sql scripts.

farnabaz (changes requested):
As comments moved to end, we can use `--` comment which helps us to improve performance a bit more

elevatebart:
Thanks @farnabaz, All good suggestions. I took them all.

I also found out that the algorithm was backwards:
I was skipping instead of inserting the records that wer already in the DB...
I fixed it and added a comment

src/runtime/internal/database.server.ts
```diff
@@ -113,20 +115,63 @@ async function _checkAndImportDatabaseIntegrity(event: H3Event, collection: stri
...
+  }
 
   await dump.reduce(async (prev: Promise<void>, sql: string) => {
     await prev
-    await db.exec(sql).catch((err: Error) => {
+
+    // in D1, there is a bug where semicolons and comments can't work together
+    // so we need to split the SQL and remove the comment
+    // @see https://github.com/cloudflare/workers-sdk/issues/3892
+    const [statement, hash] = sql.split(' -- ')
```
farnabaz, line 149: This will fail if query contains ` -- `. Imagine if the content contains `sql` snippet.

we can use `hashListFromTheDump` array to find line hash value and remove via `substring`

Something like

```ts
const hash = hashListFromTheDump[index]
const statement = sql.substring(0, sql.length - ('  -- ' + hash).length)
```

elevatebart: good catch yeh !

farnabaz (approved):
LGTM 👍 
Great work @elevatebart

Issue #3248, opened 2025-03-14
Environment requirements missing in doc


DevDengChao:
### Environment

The env I used locally:
- Operating System: Windows_NT
- Node Version:     v22.14.0
- Nuxt Version:     3.16.0
- CLI Version:      3.23.0
- Nitro Version:    2.11.6
- Package Manager:  pnpm@8.6.2
- Builder:          -
- User Config:      compatibilityDate, devtools, modules, app, experimental, routeRules, runtimeConfig
- Runtime Modules:  @unocss/nuxt@66.0.0, @vueuse/nuxt@13.0.0, vuetify-nuxt-module@0.18.3, @nuxtjs/device@3.2.4, @nuxt/content@3.3.0
- Build Modules:    -

The env I used on cloud:
- Operating System: Linux
- Node Version:     v20.10.0
- Nuxt Version:     -
- CLI Version:      3.23.0
- Nitro Version:    -
- Package Manager:  unknown
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian

### Version

v3

### Reproduction

Sorry, I cannot provide a reproduction.

### Description

I am developing an Nuxt web application locally on a Windows PC, then deploy it to Aliyun's Function Compute's NodeJS environment. 
However the official NodeJS 20 runtime is based on debian 10, which does not contain GLIBC_2.29, this makes it unable to create the contents.sqlite file when using better sqlite 3 module.

So, I'd suggest you add a environment requirement section in doc when upgrading to v3.

### Additional context

![Image](https://github.com/user-attachments/assets/e2857cd5-1b43-4669-8237-ab713e5cef2e)

### Logs

```shell-script
2025-03-14 13:39:06 Listening on http://[::]:3000 Failed to execute SQL CREATE TABLE IF NOT EXISTS _content_info (id TEXT PRIMARY KEY, "ready" BOOLEAN, "structureVersion" VARCHAR, "version" VARCHAR, "__hash__" TEXT UNIQUE); 
-- structure: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found 
(required by /code/server/node_modules/better-sqlite3/build/Release/better_sqlite3.node)
```

AloisSeckar:
[...]

PR #3261, opened 2025-03-19
Replace node:22-alpine with node:22

DevDengChao:
### 🔗 Linked issue

https://github.com/nuxt/content/issues/3248

### ❓ Type of change

### 📚 Description

node:22-alpine does not have a proper glibc, this makes better-sqlite-3 not working.
```
PS C:\Users\Admin> docker run --rm -it node:22-alpine sh
/ # ldd --version
musl libc (x86_64)
Version 1.2.5
Dynamic Program Loader
Usage: /lib/ld-musl-x86_64.so.1 [options] [--] pathname
```

### 📝 Checklist

### DIFF
--- a/docs/content/docs/6.deploy/7.docker.md
+++ b/docs/content/docs/6.deploy/7.docker.md
@@ -31,7 +31,9 @@ RUN pnpm run build
 
 # Build Stage 2
 
-FROM node:22-alpine
+# node:22-alpine is not compatible with better-sqlite-3 module
+# See https://github.com/nuxt/content/issues/3248 for more details
+FROM node:22
 WORKDIR /app
 
 # Only `.output` folder is needed from the build stage


farnabaz:
Are you sure that node-alpine does not work? I did test Content v3 with `node:22-alpine` and it was fine

I think the `better-sqlite3` bundle for Alpine is a bit different and does not require Glibc

DevDengChao:
image

I tried node:22-alpine with the examples/basic project, it does not work.

pi0:
(personally, I would suggest using glibc variants / perhaps `-slim` instead. alpine has always edge cases)

farnabaz:
@DevDengChao You are doing it wrong, When you use Alpine you must build the project inside docker. When you build it outside of Docker the CPU architecture is different and you will end up with error.

Try this:
```dockerfile
# Build Stage 1

FROM node:22-alpine AS build
WORKDIR /app
COPY package.json .npmrc ./
RUN npm i
COPY . ./
RUN npm run build

# Build Stage 2

FROM node:22-alpine
WORKDIR /app
COPY --from=build /app/.output/ ./
ENV PORT 80
ENV HOST 0.0.0.0
EXPOSE 80

CMD ["node", "/app/server/index.mjs"]
```

PR #3276, opened 2025-03-24
feat: collection hashing for multi instance support

### LINKED ISSUE #3273: Support running multiple instances on v3

Currently we are unable to reliably use Nuxt Content v3 with Nuxt Test Utils with multiple parallel workers as Nuxt Content adds content to a static set of SQL tables.

The following reproduction demonstrates the issue:
https://github.com/GerryWilko/content-sql-repro

I would like to be able to safely run Nuxt Content v3 with multiple Nuxt instances connecting to the same DB.

During module initialization all existing content tables are dropped:
https://github.com/nuxt/content/blob/96a9c4a45b3d9966f84cf15564701d9bda5cc3d5/src/module.ts#L253-L254

If we are already dropping all the content tables on module initialization we could just use a set of scoped tables within the same DB with a unique ID generated on module init. This would allow for each Nuxt instance to only interact with its own tables.

This solution may be problematic for users that are not using the built in SQLite server as they will end up with many duplicate copies of their content files. However at present if they are running multiple instances then each one is perhaps unnecessarily clearing the content dbs and they may be hitting this race condition of instances clashing with each other on start.

Alternatively we instead stop calling `dropContentTables` on module initialization and instead perform a clean-up operation of orphaned content files in each content table after content has been injected into and overwritten what is already present.

GerryWilko:
### 🔗 Linked issue

fixes #3273 

### ❓ Type of change

### 📚 Description

This PR aims to address issues related to multiple running instances of Nuxt connecting to the same collection database by optimizing the module initialization and removing the need to drop all content tables and recreate each time.

I have proposed moving the collection `hash` into the `resolveCollection` function and using the first four safe characters as part of the table name. This allows collections that have not been altered to be left untouched during initialization allowing for multiple connecting Nuxt instances to run the module without interfering with each other.

Do let me know if there is anything you would like me to change here.

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -250,14 +250,14 @@ async function processCollectionItems(nuxt: Nuxt, collections: ResolvedCollectio
   let cachedFilesCount = 0
   let parsedFilesCount = 0
 
-  // Remove all existing content collections to start with a clean state
-  db.dropContentTables()
+  // Remove all out of date collections and only create new ones
+  const { upToDateTables } = await db.dropOldContentTables(collections)
+  const newCollections = collections.filter(c => !upToDateTables.includes(c.name))
   // Create database dump
-  for await (const collection of collections) {
+  for await (const collection of newCollections) {
     if (collection.name === 'info') {
       continue
     }
-    const collectionHash = hash(collection)
     const collectionQueries = generateCollectionTableDefinition(collection, { drop: true })
       .split('\n').map(q => `${q} -- structure`)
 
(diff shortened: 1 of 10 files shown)


farnabaz:
Thanks @GerryWilko for the issue and PR,
Adding hash to the table's name is not an ideal thing to do. As a user, I don't like to see hashes in my table's name. Also what if two projects share the exact same collection?


Instead of adding hash suffix, we can add support for a custom table prefix. This way in tests each instance can define its own prefix. Also this would be useful in production, you can share sam database between two or more applications.
Instead of adding hash suffix

case-fastly:
@GerryWilko curious if you have any further thoughts on this? (per the comment above)

PR #3304, opened 2025-04-08
fix(collection): avoid double update of some record by using the hash column as index

elevatebart:
### ❓ Type of change

### 📚 Description

When two users on diferent parts of the world open the same un-deployed nuxt content powered website, they both trigger the deployment at the same time. Since it can take a moment for data to be synchronized, both can still be running the data at the same time, including the concatenation updates.

To avoid duplicating and therefore corrupting the data, I had the idea to use a temporary hash when inserting incomplete data:

example with the hash of the record being `123456798` and the column name `content`

```
INSERT INTO _col_collection VALUES ('52', '[values-slice-1...]', '123456798-50000')
UPDATE _col_collection SET content = CONCAT(content, '[values-slice-2]'), __hash__ = '123456798-100000' WHERE id = 52 AND __hash__ =  '123456798-50000'
UPDATE _col_collection SET content = CONCAT(content, '[values-slice-3]'), __hash__ = '123456798' WHERE id = 52 AND __hash__ =  '123456798-100000'
```

With this we prevent most corruption of data through multiple updates. 
We still need a way to "repair" a system that has stopped deploying mid deployment. 
But this should improve deployment stability slightly.

### 📝 Checklist

### DIFF
--- a/src/utils/collection.ts
+++ b/src/utils/collection.ts
@@ -190,7 +190,11 @@ export function generateCollectionInsert(collection: ResolvedCollection, data: P
     }
   })
 
+  console.log('values', values)
+
   const valuesHash = hash(values)
+
+  console.log('valuesHash', valuesHash)
 
   const hashColumn = opts.hashColumn !== false
 
@@ -198,7 +202,7 @@ export function generateCollectionInsert(collection: ResolvedCollection, data: P
 
   let index = 0
   const sql = `INSERT INTO ${collection.tableName} VALUES (${'?, '.repeat(valuesWithHash.length).slice(0, -2)});`
-    .replace(/\?/g, () => values[index++] as string)
+    .replace(/\?/g, () => valuesWithHash[index++] as string)
 
   if (sql.length < MAX_SQL_QUERY_SIZE) {
     return {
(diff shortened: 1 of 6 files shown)


src/utils/collection.ts
```diff
@@ -191,12 +191,11 @@ export function generateCollectionInsert(collection: ResolvedCollection, data: P
   })
 
   const valuesHash = hash(values)
-  if (opts.hashColumn !== false) {
-    values.push(`'${valuesHash}'`)
-  }
```
farnabaz, line 196: We need to add back this

src/utils/collection.ts
```diff
@@ -191,12 +191,11 @@ export function generateCollectionInsert(collection: ResolvedCollection, data: P
...
+
+  const valuesWithHash = opts.hashColumn !== false ? [...values, `'${valuesHash}'`] : values
```
farnabaz, line 195: I'm thinking of removing `hashColumn` option and add it to all, it will simplify the logic here.

I added this option to avoid adding `__hash__` to the development cache table, which causes SQL to fail. But forcing cache cleanup in the new version would be a better solution.

farnabaz (changes requested):
Thanks, @elevatebart, for the PR; it's a nice fix.
Added two change requests before

src/utils/collection.ts
elevatebart: A system with no hashColumn would not benefit from this improvement. 
I will set it up.

farnabaz (approved):
Thanks ❤️

PR #3320, opened 2025-04-21
fix: do not escape newlines

dword-design:
### ❓ Type of change

### 📚 Description

Currently newlines are escaped when putting strings as fields into @nuxt/content. That means that the newlines that are returned from e.g. `queryCollection` aren't real newlines but slashes followed by an n char. This is clearly visible when converting the field value to JSON, then the output looks something like `line1\\nline2` instead of `line1\nline2`. Concretely this breaks my [nuxt-content-body-html](https://github.com/dword-design/nuxt-content-body-html) module that generates a html for @nuxt/content posts. In the final result e.g. in code blocks, there will just be `\n` strings instead of real line breaks.

I think there are two issues:

1. The newlines are escaped when generating the inserts
2. Actual newlines are used to split the decompressed dump on client side

I've fixed them by keeping the newlines in the queries and using JSON.stringify and JSON.parse to communicate the dump between server and client instead of spliting by `\n`.

I haven't fixed the tests yet since I don't know yet if the PR gets merged.

### Reproduction

https://codesandbox.io/p/devbox/j4yj8k


farnabaz:
What issue does this PR aim to fix? Could you provide a reproduction sample?

luksak:
I just encountered this issue myself. The PR fixes the bug for me. Thanks a lot!

farnabaz:
I see
Unfortunately, not all database providers support multi-line queries. That's what module replaces new lines with `\\n`.
The issue here is not about encoding the newlines; the issue is that the module does not decode the column value when retrieving it.

What we can do is to change `the refineContentFields` function in `runtime/internal/collections.ts`, to decode the `\\n` back to `\n`.

farnabaz:
@dword-design You were right, double checked with databases, and all seems fine.

farnabaz (approved):
Thanks

PR #3332, opened 2025-04-30
docs(examples): add key to dynamic asyncData

huang-julien:
### ❓ Type of change

### 📚 Description

Async data keys should always uniquely identify a payload. Auto-keyed asyncData should be used in case of a unique payload. Data between dynamic routes are not considered as a unique payload since `queryCollection('blog').path(route.path).first()` returns a data depending on the current route.

### 📝 Checklist

### DIFF
--- a/examples/blog/app/pages/[...slug].vue
+++ b/examples/blog/app/pages/[...slug].vue
@@ -1,6 +1,6 @@
 <script lang="ts" setup>
 const route = useRoute()
-const { data: post } = await useAsyncData(() => {
+const { data: post } = await useAsyncData(computed(() => 'blog-' + route.path), () => {
   return queryCollection('blog')
     .path(route.path)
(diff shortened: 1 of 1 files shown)


examples/blog/app/pages/[...slug].vue
```diff
@@ -1,6 +1,6 @@
 <script lang="ts" setup>
 const route = useRoute()
-const { data: post } = await useAsyncData(() => {
+const { data: post } = await useAsyncData(computed(() => 'blog-' + route.path), () => {
```
Comment (farnabaz) [inline examples/blog/app/pages/[...slug].vue:3]:
```suggestion
const pageId = computed(() => 'blog-' + route.path)
const { data: post } = await useAsyncData(pageId, () => {
```

farnabaz (approved):
Thanks

PR #3349, opened 2025-05-12
chore(deps): make `better-sqlite3` dependency optional

### LINKED ISSUE #3346: Make `better-sqlite3` an optional dependency
## Environment

- `@nuxt/content` version: 3.5.1
- Bun version: v1.2.13 (64ed68c9)
- Platform: Linux 6.13.7-orbstack-00283-g9d1400e9c6 (ARM64 Docker container)

## Description

When installing `@nuxt/content` inside a Bun Docker image (ARM64, Linux), the `better-sqlite3` native module fails to build due to missing Python and a V8 symbol mismatch:

```
node: symbol lookup error: /usr/src/app/node_modules/better-sqlite3/build/Release/better_sqlite3.node: undefined symbol: _ZN2v816FunctionTemplate16InstanceTemplateEv

...

gyp ERR! find Python Python is not set from command line or npm configuration

...

Error: Could not find any Python installation to use

...

error: install script from "better-sqlite3" exited with 1
```

But Bun already provides built-in SQLite support, so requiring `better-sqlite3` as a hard dependency blocks Bun users or those without a full C++/Python toolchain.

## Proposed solution

- Move `better-sqlite3` from `"dependencies"` into `"optionalDependencies"` in `package.json`.
- Update documentation to note that `better-sqlite3` is only required for environments without native SQLite support.

I’d be happy to open a PR to implement this change if the proposed solution looks good!

dannote:
### 🔗 Linked issue

Resolves #3346

### ❓ Type of change

### 📚 Description

Make `better-sqlite3` optional to prevent native‐build failures (e.g. in Bun Docker):

- Move `better-sqlite3` into `peerDependencies` (marked optional)
- Add runtime checks so it’s only required when actually used

### 📝 Checklist


farnabaz:
Thanks for the PR
Could you merge main into your PR and update the installation docs for each environment? https://content.nuxt.com/docs/getting-started/installation

It would be nice also to have a message to note about moving better-sqlite as an optional dependency.

Ideally, I would be happy if we could prompt the user and ask her permission to install `better-sqlite` automatically. Just like what Nuxt does for `@nuxt/image` module if it's missing.

- https://github.com/nuxt/nuxt/blob/76b300eddae6d755948239c305bfcc587dd74197/packages/nuxt/src/components/plugins/loader.ts#L58-L65
- https://github.com/nuxt/nuxt/blob/76b300eddae6d755948239c305bfcc587dd74197/packages/nuxt/src/core/features.ts

farnabaz (approved):
Thanks

PR #3390, opened 2025-06-03
fix: exclude `node_modules` from watching

pi0:
### ❓ Type of change

### 📚 Description

Currently [undocs](https://github.com/unjs/undocs) based documentations fail with `EMFILE: too many open files, watch` on macOS because the whole `node_modules` and `.nuxt` dirs are being watched. 

I think a proper long-term fix would be respecting `source.exclude` (like [this](https://github.com/unjs/undocs/blob/main/app/content.config.ts#L10)) with separate watchers, but as a quick remedy, I made this PR.

### 📝 Checklist


farnabaz:
What if users want to load data from a package? Hardcoding `node_modules` will be a breaking change.

The ignore list can be created by merging `source.exclude` items with `source.cwd` and passing it to the watcher.

pi0:
Yes as mentioned my primary goal was quick workaround and sharing issue. Feel free to close my PR or update it with probper exclude support. (we need to resolve globs as latest chokidar dropped support)

pi0:
thnx;)

PR #3421, opened 2025-06-20
fix(dev): dont ignore subfolders

hendrikheil:
### ❓ Type of change

### 📚 Description

Currently, for subfolders, path would end up being shorter than prefix, which causes it to never match. This leads to all content in those folders being ignored.

### 📝 Checklist

### DIFF
--- a/src/utils/dev.ts
+++ b/src/utils/dev.ts
@@ -102,7 +102,9 @@ export async function watchContents(nuxt: Nuxt, options: ModuleOptions, manifest
     return []
   }).filter(({ source }) => source.cwd)
 
-  const dirsToWatch = Array.from(new Set(sourceMap.map(({ prefix }) => prefix))).filter(Boolean)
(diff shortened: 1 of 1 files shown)


src/utils/dev.ts
```diff
@@ -102,9 +102,7 @@ export async function watchContents(nuxt: Nuxt, options: ModuleOptions, manifest
     return []
   }).filter(({ source }) => source.cwd)
 
-  const dirsToWatch = Array.from(new Set(sourceMap.map(({ source }) => source.cwd)))
-    // Filter out empty cwd for custom collections
-    .filter(Boolean)
+  const dirsToWatch = Array.from(new Set(sourceMap.map(({ prefix }) => prefix))).filter(Boolean)
```
farnabaz, line 105:
```suggestion
  const dirsToWatch = Array.from(new Set(sourceMap.map(({ prefix }) => prefix)))
    // Filter out empty cwd for custom collections
    .filter(Boolean)
```

PR #3430, opened 2025-06-26
fix(docs): update query method to use `first()` for consistency

cirolosapio:
### ❓ Type of change

there are many [queryContent](https://github.com/nuxt/content/blob/main/playground/content/1.real-content/content.md?plain=1#L41) calls using various methods that typeScript complains about in the [playground](https://github.com/nuxt/content/blob/main/playground/content/1.real-content/content.md?plain=1#L71) folder as well
should those be updated too? I'm new to nuxt content


farnabaz:
Thanks for the PR
You can ignore playground for now, I have plan to rewrite it

farnabaz (approved):
Thanks

PR #3478, opened 2025-07-24
fix: improve websocket and watcher cleanup handling

productdevbook:
## Summary
- Add null checks before closing websocket and listener to prevent errors
- Use `listener.close()` instead of `listener.server.close()` for proper cleanup
- Remove all listeners from watcher before closing to prevent memory leaks

## Changes
- Added conditional checks for websocket and listener existence before closing
- Fixed listener close method call
- Added `removeAllListeners()` call for watcher cleanup
- Added null check for watcher in close hook

## Test plan
- [ ] Verify that the development server starts normally
- [ ] Confirm that websocket connections work as expected
- [ ] Test that the server closes cleanly without errors when stopping
- [ ] Ensure no memory leaks occur during server restarts

### DIFF
--- a/src/utils/dev.ts
+++ b/src/utils/dev.ts
@@ -43,8 +43,13 @@ export async function startSocketServer(nuxt: Nuxt, options: ModuleOptions, mani
 
     nuxt.hook('close', async () => {
       // Close WebSocket server
-      await websocket?.close()
-      await listener?.server?.close()
+      if (websocket) {
+        await websocket.close()
+      }
+      // Close listener server
+      if (listener) {
+        await listener.close()
+      }
     })
   }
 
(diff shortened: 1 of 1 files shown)


src/utils/dev.ts
```diff
@@ -211,8 +216,10 @@ export async function watchContents(nuxt: Nuxt, options: ModuleOptions, manifest
   }
 
   nuxt.hook('close', async () => {
-    watcher.close()
-    db.close()
+    if (watcher) {
+      watcher.removeAllListeners()
+      watcher.close()
```
productdevbook, line 221:
```suggestion
      watcher.close()
      db.close()
```

PR #3479, opened 2025-07-24
docs: i18n example useless import

edimitchel:
### ❓ Type of change

### 📚 Description

Useless import of localeProperties

PS: Thank you for that example which comes a bit late but I know will help a lot for others!

That said, I still think this way is very difficult to use in long term when content evolves.. We systematically need to update other content.. I found a solution by using AI which creates PR of the content translated in order to review. I think we could improve the UI to help the content editor to switch between languages..

### DIFF
--- a/docs/content/docs/7.integrations/01.i18n.md
+++ b/docs/content/docs/7.integrations/01.i18n.md
@@ -81,7 +81,7 @@ import { withLeadingSlash } from 'ufo'
 import type { Collections } from '@nuxt/content'
 
 const route = useRoute()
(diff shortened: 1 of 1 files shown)


farnabaz:
> That said, I still think this way is very difficult to use in long term when content evolves.. We systematically need to update other content.. I found a solution by using AI which creates PR of the content translated in order to review. I think we could improve the UI to help the content editor to switch between languages..

Feel free to open an issue to discuss it

edimitchel:
There it is :)
https://github.com/nuxt/content/issues/3492

PR #3500, opened 2025-08-08
fix: use listhen's `publicUrl` if available

farnabaz:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/runtime/internal/websocket.ts
+++ b/src/runtime/internal/websocket.ts
@@ -82,12 +82,6 @@ export function useContentWebSocket() {
     // WebSocket Base URL
     const wsURL = new URL(joinURL((useRuntimeConfig().public.content as { wsUrl: string }).wsUrl, 'ws'))
     wsURL.protocol = window.location.protocol === 'https:' ? 'wss:' : 'ws:'
-    if (wsURL.hostname === 'localhost') {
-      // If the hostname is localhost, use the hostname of the current window
-      // This is useful for local development with a reverse proxy
-      // As alternative, you can use the `publicURL` option to set the public URL of the WebSocket server
-      wsURL.hostname = window.location.hostname
-    }
 
     logger.log(`WS connect to ${wsURL}`)
 
--- a/src/runtime/internal/websocket.ts
+++ b/src/runtime/internal/websocket.ts
@@ -1,5 +1,6 @@
 import { loadDatabaseAdapter } from './database.client'
 import { useRuntimeConfig, refreshNuxtData } from '#imports'
+import { joinURL } from 'ufo'
 
 const logger = {
   log: (...args: unknown[]) => console.log('[Content]', ...args),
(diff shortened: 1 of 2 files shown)


hendrikheil:
Great work! I verified this works, you just have to specify `watch.port` and `watch.publicUrl` where `publicUrl: wss://{domain}`

farnabaz:
@dargmuesli I think this PR also affect/help you, but it should not break the current behavior.

src/runtime/internal/websocket.ts
```diff
@@ -79,9 +80,14 @@ export function useContentWebSocket() {
...
-    const wsURL = new URL(`${(useRuntimeConfig().public.content as { wsUrl: string }).wsUrl}ws`)
+    const wsURL = new URL(joinURL((useRuntimeConfig().public.content as { wsUrl: string }).wsUrl, 'ws'))
     wsURL.protocol = window.location.protocol === 'https:' ? 'wss:' : 'ws:'
-    wsURL.hostname = window.location.hostname
+    if (wsURL.hostname === 'localhost') {
+      // If the hostname is localhost, use the hostname of the current window
+      // This is useful for local development with a reverse proxy
+      // As alternative, you can use the `publicURL` option to set the public URL of the WebSocket server
+      wsURL.hostname = window.location.hostname
+    }
```
maximepvrt, line 90: For me, this code should be removed. If we specifically want to force `localhost` via the config `content.watch.publicURL = 'https://localhost'`, it gets replaced by the current route. In this case, we should use `publicURL`.

src/utils/dev.ts
```diff
@@ -36,7 +36,8 @@ export async function startSocketServer(nuxt: Nuxt, options: ModuleOptions, mani
       listener = await listen(() => 'Nuxt Content', websocketOptions)
 
       // Register ws url
-      ;(nitro.options.runtimeConfig.public.content as Record<string, unknown>).wsUrl = listener.url.replace('http', 'ws')
+      const publicConfig = nitro.options.runtimeConfig.public.content as Record<string, unknown>
+      publicConfig.wsUrl = (websocketOptions.publicURL || listener.url).replace('http', 'ws')
```
maximepvrt, line 40: `publicConfig.wsUrl = (websocketOptions.publicURL || process.env.publicURL || listener.url).replace('http', 'ws')`

to use option `--publicURL` defined by the `nuxt dev` command.

maximepvrt (changes requested):

src/runtime/internal/websocket.ts
maximepvrt: I don’t think browsers block connecting to `localhost` from another domain. As far as I know, WebSockets don’t have CORS validation in the same way as HTTP requests.

Ref: [StackOverflow – Why is there no same origin policy for WebSockets?](https://stackoverflow.com/questions/23674199/why-is-there-no-same-origin-policy-for-websockets-why-can-i-connect-to-ws-loc)

maximepvrt: Before this commit (https://github.com/nuxt/content/pull/3344/files
), with a proxy the hostname was always localhost by default, and for me it worked fine.

maximepvrt:
[Laravel Valet proxies](https://laravel.com/docs/12.x/valet#proxying-services) allow binding a domain to a service running on a specific port, but they don’t support handling multiple services with different ports.

Example:

```
monsite.test    => https://localhost:3000
monsite.test:4000 => https://localhost:4000 (not supported)
```

That’s why I’d rather keep using `https://localhost:4000` directly for **Hot Content Reload**, to avoid **CORS issues** with websockets.

src/runtime/internal/websocket.ts
farnabaz: Makes sense, thanks

PR #3543, opened 2025-09-17
Fix: Typo in Dockerfile lock file name

emielmolenaar:
### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Fixes a typo in a Dockerfile example.

PR #3545, opened 2025-09-17
Fix: Update COPY command in docs to support multiple lock files

emielmolenaar:
### ❓ Type of change

### 📚 Description

See #3543.  Bun v1.2 changed the default lockfile format to the text-based bun.lock. Existing binary bun.lockb lockfiles can be migrated to the new format by running bun install --save-text-lockfile --frozen-lockfile --lockfile-only and deleting bun.lockb.


emielmolenaar:
Thanks!

PR #3587, opened 2025-10-27
docs: Clarify Cloudflare Pages and D1

SolidKalium:
Emphasize the necessity of creating a D1 database and how to configure wrangler so `nuxi preview` works.

### ❓ Type of change

#### Changes

- Places more emphasis on a D1 database being required
- Provides a minimal example wrangler config for running preview locally
- Explains what this example config does
- Links to Nuxt Content's docs section for D1 configuration
- Links more directly to Cloudflare's documentation about creating and binding a D1 database

#### Why

My initial reading of the statement that the module will "prepare the necessary configuration" caused me to misunderstand the importance of D1 being required. I think the new emphasis should help prevent that misreading.

I also had trouble running `nuxi preview` locally to check a mismatch between local and prod behavior. I no longer recall if `nuxi preview` actually helped me solve my problems, but I suspect other people probably don't want to be stuck configuring preview when they think it will help debugging something else.

For context, the issues I was debugging (and have since resolved) were:
- Some content not loading correctly. This was likely because another source told me to use generate instead of build, and I then also didn't have D1 set up.
- An infinite 404 loop. This was likely caused by the fact that I didn't have an error.vue file and the fact that my default slug was throwing a 404 when content was missing. But local dev was ok because it provides an error page.
- A file I created postbuild wasn't being added to _routes, because _routes is created during build and I didn't understand that part of Cloudflare enough to expect the issue.

### 📝 Checklist

### DIFF
--- a/docs/content/docs/6.deploy/4.cloudflare-pages.md
+++ b/docs/content/docs/6.deploy/4.cloudflare-pages.md
@@ -28,11 +28,13 @@ export default defineNuxtConfig({
 ```
 
 ## D1 Database
+
 A D1 database connected to the app is **required** for the Nuxt Content module to work. By default the module uses the binding name **`DB`**. You can override the [database configuration](/docs/getting-started/configuration#d1) by providing your own in `nuxt.config.ts`.
 
 After creating a new Cloudflare Pages project, you need to create a new D1 database and connect it to the project. Make sure to use the same binding name as the module is using.
 
 ### Local Preview
+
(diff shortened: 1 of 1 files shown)


docs/content/docs/6.deploy/4.cloudflare-pages.md
```diff
@@ -7,35 +7,51 @@ description: Deploy your Content app to Cloudflare Pages
...
+```
 
 That's it :tada:
 
-Checkout:
+Relevant resources:
 
 - [Nuxt Deploy documentation](https://nuxt.com/deploy/cloudflare)
 - [Cloudflare D1 documentation](https://developers.cloudflare.com/d1/)
+  - [Create and bind a D1 database](https://developers.cloudflare.com/d1/get-started/)
```
atinux:
```suggestion
- [Create and bind a D1 database](https://developers.cloudflare.com/d1/get-started/)
```

atinux:
Thank you for your help, would you be happy to also update the Cloudflare Workers page @SolidKalium ?

SolidKalium:
I haven't built for that target directly and don't have a plan to. So I don't think I'd be a good author for that. But feel free to copy over any relevant content I added.

PR #3591, opened 2025-10-31
perf: replace internal object records with maps

OrbisK:
### ❓ Type of change

### 📚 Description

I was actually looking for other things and came across these improvements. In some places, you could certainly use a map instead of an object record, but that would then affect some exported APIs.

### 📝 Checklist


farnabaz:
Thanks for the PR,
Although that PR looks good, but I'm not sure about the performance improvement here, `Map` has overhead of indexing had is good choice for large set of data, but here we are using `object` for small set of data which is slightly faster in object as far as I know.

It would be great if we can have a benchmark two decide between them.

farnabaz (approved):
I ran a small test on Map and Object and the result was almost same with small data.
Beside performance using map is better for security and readability.

Thanks for PR

PR #3597, opened 2025-11-04
perf: use `moduleDependencies` to install mdc module

danielroe:
### ❓ Type of change

### 📚 Description

this uses the new module dependencies feature in https://github.com/nuxt/nuxt/releases/tag/v4.1.0 to specify options for other modules (and require them).

without this, the `installModule` call will overwrite any module dependencies that are specified by other modules - so this PR is needed for https://github.com/nuxt/ui/pull/5384.

**Note**: I would recommend instead of passing options from content -> mdc, that we have a single place to define them (ie. just use mdc instead), as it can otherwise create non-optimal dependency chains... and I worry bugs will appear

could we move/document `content.renderer.alias` -> `mdc.components.map` instead, and so on?

### 📝 Checklist


farnabaz (approved):
Thanks

PR #3605, opened 2025-11-06
docs(markdown): add info to the complete MDC docs

maximepvrt:
Every time, I look for this link to find the full MDC syntax documentation 😅

For example, the current content markdown page doesn’t explain inline blocks, nested blocks, etc.

And it’s this page that is currently given as the MDC doc here: https://github.com/nuxt-content/mdc?tab=readme-ov-file#features

Maybe it would be better to create a dedicated documentation page for the MDC syntax on the Nuxt Content site, which would be as complete as possible, rather than spreading the documentation across multiple places as it is currently.

### ❓ Type of change

### 📝 Checklist


farnabaz:
Let's keep the link for now. We will create dedicated page for MDC Syntax and Spec soon

PR #3636, opened 2025-12-09
feat(search): add minHeading/maxHeading options

onmax:
Adds `minHeading` and `maxHeading` options to filter which heading levels create new sections.

```ts
queryCollectionSearchSections('docs', { minHeading: 'h2', maxHeading: 'h3' })
```

Default: `h1` to `h6` (backward compatible)

### DIFF
--- a/docs/content/blog/ui-pro-docs-migration.md
+++ b/docs/content/blog/ui-pro-docs-migration.md
@@ -229,8 +229,6 @@ export default defineNuxtConfig({
 })
 ```
 
-Finally, in order to keep the [app config file updatable](/docs/studio/config) from Studio, we just need to update the helper import of the `nuxt.schema.ts` file from `@nuxthq/studio/theme` to `@nuxt/content/preview`.
-
 ::prose-tip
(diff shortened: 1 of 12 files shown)


farnabaz:
I like the idea about separator but I'm not sure how useful it could be because by default other elements do not have `id` attribute. The only way to achieve it for `hr` is
```
Paragraph

::hr{#my-id}
::

Paragraph
```

onmax:
Good point. In my personal needs, the main use case is customising which heading levels trigger a new section (e.g. 'h2' or 'h3' rather than all headings).

Adding support for any component, like `hr`, is just an added bonus, providing flexibility; users who want it can add IDs via MDC syntax.

Should I add a note about this in the documentation, or should we only allow headings in the separator option?

farnabaz:
I think headers should be enough. Let’s face it: if you want to create a section, you need a title. What’s the purpose of a section without a title and proper link? In my view, if it doesn’t have a title, it’s most likely a continuation of the previous section.

So I'll go with headers only option.

onmax:
updated the pr. changed api from `separators: string[]` to `minHeading/maxHeading`.

the array approach allowed weird combos like `['h2', 'h4']` skipping h3. range-based is cleaner and self-documenting - you know exactly what headings will split sections

Issue #3672, opened 2026-01-05
Expose collections config & schemas via a `useContent` composable


maximepvrt:
While working on validating field formats against the generated **JSON Schema**, I found it difficult to **access or inspect the schema associated with a collection**.

It would be very helpful to have a composable such as `useContent` that could return:

* the list of available **collections**
* the **JSON Schema** associated with each collection

This would greatly simplify:

* schema validation during development
* testing edge cases (e.g. `date` vs `datetime`)
* building custom tooling on top of Nuxt Content / Studio

#### What I tried

I attempted to access the collections directly using:

```ts
import { collections } from '#content/preview'
```

but without success, this resulted in an error and didn’t seem to be supported in this context.

maximepvrt:
@farnabaz can you reopen ?

maximepvrt:
@farnabaz can you reopen ?

PR #3675, opened 2026-01-07
fix: allow hidden property in editor without redefine validation #3661

maximepvrt:
@farnabaz I realized I forgot to commit everything in my PR #3661. With this update, I was wondering if we could also simplify the usage when it’s combined with `inherit`?

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/utils/schema/index.ts
+++ b/src/utils/schema/index.ts
@@ -14,7 +14,7 @@ type Property<T> = T & {
 export function property<T extends ContentStandardSchemaV1>(input?: T): Property<T> {
   const $content = {} as ContentConfig
 
-  const attachContent = (schema: ContentStandardSchemaV1) => {
+  const attachContent = (schema: ContentStandardSchemaV1 | null) => {
     if (!schema) return
     const vendor = detectSchemaVendor(schema)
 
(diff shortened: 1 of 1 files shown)


farnabaz:
Do you mean using inherit with `property()` like `property().inherit(....)`? Or you have another proposal?

maximepvrt:
@farnabaz Yes, exactly ! the goal is to avoid having to write `z.object({})` when there’s nothing to define, and just use `property().editor()` or `property().inherit()` directly.

farnabaz:
@maximepvrt have you already tested and it worked for you?

maximepvrt:
@farnabaz The tests seem to pass on my side, but I’d appreciate it if you could confirm it’s all good, as I have some doubts. It’s quite tricky to validate the JSON Schema output at the moment, and there’s not really a place in the playground to see the generated JSON. That’s also why I suggested adding this composable: https://github.com/nuxt/content/issues/3672

farnabaz:
When input is optional in `property()` we will lose the return type, which lead to type issue. 
Screenshot 2026-02-18 at 12 31 42


Another issue, I think we already have this: 

```
// try attaching $content context to input (input is null -> no context)
attachContent(input || null)

// return proxy around input without attached context
return createProxy(input || {})
  ```

maximepvrt:
@farnabaz You're right, this is definitely more complex than my initial approach.

Ideally, the property function should be able to retrieve the validation format from `detectSchemaVendor `(used in `defineCollection`), but I'm struggling to bridge that gap since `property` runs independently.

I suspect we might need to temporarily store the metadata provided by `property` and then inject it into the model during the `defineCollection` execution. However, I'm not quite sure how to implement this cleanly.

PR #3673, opened 2026-01-05
feat: cast date field from datetime to date string format

maximepvrt:
This change was made **in anticipation of https://github.com/nuxt-content/nuxt-studio/issues/163**, which will introduce proper support for `datetime` inputs.
Until then, the affected field is now **cast to a `date` string format** to avoid ambiguity and ensure compatibility with the current behavior.

Testing this change was **fairly cumbersome**, especially when validating the expected **JSON Schema format**.
It would be very helpful to be able to **easily retrieve or inspect the generated schema** to simplify this process.
I’ve opened a **dedicated issue** about this here: https://github.com/nuxt/content/issues/3672.

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #3698, opened 2026-01-24
fix: `syntax error` with datetime fields and standardize date/datetime formats

maximepvrt:
This PR addresses two separate improvements:

1. **Fix SQL syntax error** introduced by https://github.com/nuxt/content/pull/3673/changes

```
 ERROR  [unhandledRejection] near "20": syntax error                  14:55:45

    at Database.exec (node_modules/.pnpm/better-sqlite3@12.6.0/node_modules/better-sqlite3/lib/methods/wrappers.js:9:14)
    at Object.exec (node_modules/.pnpm/db0@0.3.4_@electric-sql+pglite@0.3.15_@libsql+client@0.17.0_encoding@0.1.13__better-sql_71033862cb6d26be682bfee7871b05d1/node_modules/db0/dist/connectors/better-sqlite3.mjs:24:26)
    at Object.exec (src/utils/database.ts:156:10)
    at processCollectionItems (src/module.ts:423:10)
    at async src/module.ts:252:20
    at async initNuxt (node_modules/.pnpm/nuxt@4.2.2_91645c95ed588071e154b0779b0d0c3a/node_modules/nuxt/dist/index.mjs:5390:3)
    at async #initializeNuxt (node_modules/.pnpm/@nuxt+cli@3.32.0_cac@6.7.14_magicast@0.5.1/node_modules/@nuxt/cli/dist/dev-KB30iboK.mjs:492:3)
    at async NuxtDevServer.init (node_modules/.pnpm/@nuxt+cli@3.32.0_cac@6.7.14_magicast@0.5.1/node_modules/@nuxt/cli/dist/dev-KB30iboK.mjs:384:3)
    at async initialize (node_modules/.pnpm/@nuxt+cli@3.32.0_cac@6.7.14_magicast@0.5.1/node_modules/@nuxt/cli/dist/dev-KB30iboK.mjs:685:2)
    at async Object.run (node_modules/.pnpm/@nuxt+cli@3.32.0_cac@6.7.14_magicast@0.5.1/node_modules/@nuxt/cli/dist/dev-CsAvgRwH.mjs:205:51)
```

2. **Standardize date/datetime formats**

   - Dates are now stored as `YYYY-MM-DD`
   - Datetimes are now stored as `YYYY-MM-DD HH:mm:ss`

These are the default storage formats for dates in databases, ensuring consistent behavior across queries.

### DIFF
--- a/src/utils/content/transformers/utils.ts
+++ b/src/utils/content/transformers/utils.ts
@@ -6,6 +6,10 @@ export const defineTransformer = (transformer: ContentTransformer) => {
 
 export const formatDateTime = (datetime: string): string => {
   const d = new Date(datetime)
+  if (Number.isNaN(d.getTime())) {
+    throw new TypeError(`Invalid datetime value: "${datetime}"`)
+  }
+
   const hours = d.getHours()
   const minutes = d.getMinutes()
   const seconds = d.getSeconds()
@@ -15,6 +19,10 @@ export const formatDateTime = (datetime: string): string => {
 
 export const formatDate = (date: string): string => {
   const d = new Date(date)
+  if (Number.isNaN(d.getTime())) {
+    throw new TypeError(`Invalid date value: "${date}"`)
+  }
+
   const year = d.getFullYear()
   const month = d.getMonth() + 1
   const day = d.getDate()


maximepvrt:
@joehoel This PR should resolve the issue, I would appreciate if you can verify using `npm i https://pkg.pr.new/@nuxt/content@3698`

maximepvrt:
@farnabaz If you prefer using an ISO datetime format like `YYYY-MM-DDTHH:mm:ss` (with the `T` separating date and time), you’ll need to transform the values returned by Postgres for datetime fields. Otherwise, Nuxt Studio will generate a conflict if the format doesn’t match the entered value (Postgres removes the `T` and replaces it with a space).

I can also look into removing the dependency on `@internationalized/date` if you want, I found a solution using the native Date API, it’s just a bit more verbose.

farnabaz:
Thanks for the PR @maximepvrt,
Well I would prefer to remove `@internationalized/date` as dependency. It's install size is ~1Mb and we are using it for small part of application, IMO it does not worth it if we have a another solution with less overhead.
What is your opinion on it?

maximepvrt:
> @farnabaz If you prefer using an ISO datetime format like `YYYY-MM-DDTHH:mm:ss` (with the `T` separating date and time), you’ll need to transform the values returned by Postgres for datetime fields. Otherwise, Nuxt Studio will generate a conflict if the format doesn’t match the entered value (Postgres removes the `T` and replaces it with a space).

@farnabaz Could you clarify which datetime format we should officially validate for Nuxt Content? Should we use ISO with the T separator? If so, we’ll need a transformer for Postgres timestamps to produce this format, but I’m not sure how to implement it.

farnabaz:
> Could you clarify which datetime format we should officially validate for Nuxt Content? Should we use ISO with the T separator? If so, we’ll need a transformer for Postgres timestamps to produce this format, but I’m not sure how to implement it. @maximepvrt 

I think your current implementation is great, and  `YYYY-MM-DD HH:mm:ss` looks good. I think you already resolve your this issue by choosing  `YYYY-MM-DD HH:mm:ss` format.
Am I missing something here?

maximepvrt:
It's perfect ! Just making sure we're aligned on the format

Issue #3695, opened 2026-01-22
`no such table`: error in production since v3.11.0 (works in dev)


maximepvrt:
### Environment

│  Operating system      macOS 25.2.0                                                                                      │
│ │  CPU                   Apple M3 Pro (11 cores)                                                                           │
│ │  Node.js version       v22.22.0                                                                                          │
│ │  nuxt/cli version      3.32.0                                                                                            │
│ │  Package manager       pnpm 10.28.1                                                                                      │
│ │  Nuxt version          4.2.2                                                                                             │
│ │  Nitro version         2.13.1                                                                                            │
│ │  Builder               vite 7.3.1                                                                                        │
│ │  Config                $development, compatibilityDate, content, extends, hooks, nitro, routeRules, runtimeConfig, site  │
│ │  Modules               -

### Version

v3.11.0

### Reproduction

https://github.com/benevolt/ligilab-front/tree/bug-content-3.11/benevolt (branch `bug-content-3.11`)

### Description

Since upgrading to **Nuxt Content v3.11.0**, we are experiencing an error **only in production**, while:

* everything works correctly in **development**
* it was working perfectly in **production with the previous version**

The error occurs when calling Nuxt Content’s internal query endpoint.

```
[request error] [unhandled] [POST] http://localhost:3000/__nuxt_content/blogEngagement/query?v=v3.5.0--u00c6T4vV1FHY8bnnmoo5FSsOD4G-c8yUAI68N9RwAE
 H3Error: no such table: _content_blogEngagement
    at StatementWrapper._statement (file:///Users/maxime/Repos/ligilab-front/benevolt/.output/server/chunks/nitro/nitro.mjs:18399:56)
    ... 8 lines matching cause stack trace ...
    at async C$1 (file:///Users/maxime/Repos/ligilab-front/benevolt/.output/server/chunks/nitro/nitro.mjs:915:7329) {
  cause: Error: no such table: _content_blogEngagement
      at StatementWrapper._statement (file:///Users/maxime/Repos/ligilab-front/benevolt/.output/server/chunks/nitro/nitro.mjs:18399:56)
      at StatementWrapper.all (file:///Users/maxime/Repos/ligilab-front/benevolt/.output/server/chunks/nitro/nitro.mjs:18408:21)
      at Object.all (file:///Users/maxime/Repos/ligilab-front/benevolt/.output/server/chunks/nitro/nitro.mjs:18446:30)
[...]

PR #3701, opened 2026-01-25
fix(api): do not forward accept encoding header

dargmuesli:
### 🔗 Linked issue

Resolves https://github.com/nuxt/content/issues/3695

### ❓ Type of change

### 📚 Description

Forwarding ~~all headers~~ the `Accept-Encoding` header can make the fetch request return data with brotli compression which is not expected at the moment. Simple fix is to revert the change to headers used.

~~I've added an additional change that adds support for Nuxt CSRF protection as provided by the security module which uses the csurf module internally. This was the original reason for allowing more headers. In the future we may add a module option to whitelist a certain set of headers instead, but that's something to discuss in the future. If this feature addition is not wanted, simply remove the second line!~~

### 📝 Checklist

### DIFF
--- a/src/runtime/internal/api.ts
+++ b/src/runtime/internal/api.ts
@@ -8,6 +8,8 @@ async function fetchContent<T>(
   options: NonNullable<Parameters<typeof $fetch>[1]>,
 ): Promise<T> {
   const headers = event ? getRequestHeaders(event) : {}
+  headers['accept-encoding'] = undefined // prevent unsupported enconding issue (https://github.com/nuxt/content/pull/3701)
+
   const url = `/__nuxt_content/${collection}/${path}`
   const fetchOptions = {
     ...options,


farnabaz:
> Forwarding all headers can make the fetch request return data with brotli compression which is not expected at the moment.

@dargmuesli  Can't we just enforce `Accept-Encoding` to `gzip, deflate` or dropping that header? If it's only about `brotli`, we can simply drop the header!

obj63mc:
@farnabaz - I have tested simply overwriting the header of Accept-Encoding in the nuxt.config routeRules but when running in production mode, since v3.11.x whenever using server side rendering, the following errors are still always returned.  I even tried removing the header completely - 

Route Rules:
```
routeRules:{
		'/__nuxt_content/**':{
			headers: {
				'Accept-Encoding': 'gzip, deflate' // or just '' to remove
			}
		}
}
```
Errors:
```
Database integrity check failed TypeError
 code: 'Z_DATA_ERROR',
  [cause]: Error: incorrect header check
...
[request error] [unhandled] [POST] http://localhost:3000/__nuxt_content/{contenttype}/query?v=v3.5.0--xAjoPoXoSfYkJiTVHdgWAiZ4y8fccTFopoR-DtcwS4s
 SqliteError: no such table: _content_...
```

As a note frontend once the page is loaded and navigating nuxt content works as expected but SSR is still broken

farnabaz:
@obj63mc Could you try to monkey patch in node modules and test?

obj63mc:
So testing this - as the code is set in v3.11.x - when passing all headers the following are passed -

```
 accept: 'text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7',
'accept-encoding': 'gzip, deflate, br, zstd'
```

Using the fix in this PR and only passing cookie header the headers are 

```
accept: 'application/json'
```

If I hard code to delete the accept-encoding header just after the creation of the fetchOptions [runtime/internal/api.js with](https://github.com/nuxt/content/blob/b3fa02598152967209de75328ec3c073844a53b5/src/runtime/internal/api.ts#L12C1-L19C4) - 
```
delete(fetchOptions.headers['accept-encoding']);
```

then this functions as it should. 

I also tried just overwriting the header in the nuxt.config routeRules but that would not work and is being ignored.
So my recommendation is if you simply want to pass all headers we either need to manually remove the accept-encoding header or set it to something that will work like - 
```
'accept-encoding': '*'
```

dargmuesli:
I'm currently testing another fix related to compression, will update you when I got the results and check the solution proposed above too of course.

tux2nicolae:
Are there any workarounds until a proper fix is made?

obj63mc:
I currently have not found a way to work around this other than to run version 3.10 or by editing the internal/api.js file manually in node modules.

dargmuesli:
You can [pnpm patch](https://pnpm.io/cli/patch) dependencies, I'm sure there is similar functionality for other package managers. See my patch here: https://github.com/maevsi/vibetype/blob/372ce28eacbb646bdbaa2135cdfbb827eae72baf/patches/%40nuxt__content.patch

danielwaltz:
Just FYI for anyone here who needs it before this gets resolved, as a possible workaround you can also disable the `nitro.compressPublicAssets` option if you have it enabled.

Of course this means that all compression will be disabled in production for you, so I'd only recommend this workaround if you have a low traffic site!

PR #3703, opened 2026-01-26
docs(i18n): Handle slug as array in computed property

mmeester:
### ❓ Type of change

### 📚 Description

The example in the code for i18n only handles slugs that have a single level, for slugs with multiple levels like `/nl/foo/bar`, `route.params.slug` will be `['foo', 'bar']` resulting in foo,bar when transforming it to a string.

### 📝 Checklist


farnabaz (approved):
Thanks

PR #3704, opened 2026-01-26
fix(ssr): local content components on Cloudflare Workers

onmax:
## Summary
- Restore lazy loading for local content components on Cloudflare Workers by introducing `localComponentLoaders`.
- Avoid runtime `import('#content/components')` in `ContentRenderer`.
- Fail fast with a clear error when a named export is missing.

## Why
Cloudflare Workers SSR fails to render local MDC components when `#content/components` is dynamically imported. Loading local components via per-component async loaders avoids the Worker import issue while keeping SSR intact.

## Previews (repro on same worker)
- Main (released `@nuxt/content@3.11.0`) **fails SSR** for local components: [preview](https://content-ssr-bug-preview-1770305752-content-ssr-check-1770301164.maximogarciamtnez.workers.dev)
  - SSR HTML does **not** include `SSR: Component rendered on server`.
- This PR **fixes SSR**: [preview](https://content-ssr-fix-preview-1770306223-content-ssr-check-1770301164.maximogarciamtnez.workers.dev)
  - SSR HTML **does** include `SSR: Component rendered on server`.

## Tests
- Manual CF Workers preview SSR check (main vs PR, same worker + D1).
- Manual dev check with `@nuxtjs/seo` enabled (no missing-export errors).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [3.11.2](https://github.com/nuxt/content/compare/v3.11.1...v3.11.2) (2026-02-05)
+
+### Bug Fixes
+
+* **studio:** fallback to env variable to detect ai feature ([#3713](https://github.com/nuxt/content/issues/3713)) ([3fc8b7b](https://github.com/nuxt/content/commit/3fc8b7b06fa45136d48258e64d49cca1961c313d))
+
 ## [3.11.1](https://github.com/nuxt/content/compare/v3.11.0...v3.11.1) (2026-02-05)
 
 ### Features
--- a/docs/content/docs/4.utils/4.query-collection-search-sections.md
+++ b/docs/content/docs/4.utils/4.query-collection-search-sections.md
@@ -23,7 +23,7 @@ The `queryCollectionSearchSections` utility is available in both Vue and Nitro. 
 ## Type
 
 ```ts
-function queryCollectionSearchSections(collection: keyof Collections, opts?: { ignoredTags?: string[], minHeading?: string, maxHeading?: string }): ChainablePromise<T, Section[]>
+function queryCollectionSearchSections<T extends keyof PageCollections>(collection: T, opts?: { ignoredTags?: string[], minHeading?: string, maxHeading?: string, extraFields?: Array<keyof PageCollections[T]> }): ChainablePromise<T, Section[]>
 
 interface ChainablePromise<T extends keyof PageCollections, R> extends Promise<R> {
   where(field: keyof PageCollections[T] | string, operator: SQLOperator, value?: unknown): ChainablePromise<T, R>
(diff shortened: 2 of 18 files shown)


mateusznarowski:
Now, the application built with the `cloudflare_module` preset renders correctly on the server side (so that part can be considered fixed), but an error appears in `dev` on the client side, the page flickers, and a 500 error appears with the following message:

```
The requested module '/_nuxt/@fs/Users/mateusznarowski/temp/content-issue/node_modules/.pnpm/nuxt@4.3.0_@parcel+watcher@2.5.6_@vue+compiler-sfc@3.5.27_better-sqlite3@12.6.2_cac@6.7_d2aeb2292c1af4998397e345ba461360/node_modules/nuxt/dist/app/components/nuxt-stubs.js?v=9bbd892e' does not provide an export named 'default'
```

`Nuxt DevTools`
```
fs/Users/mateusznarowski/temp/content-issue/node_modules/.pnpm/nuxt@4.3.0_@parcel+watcher@2.5.6_@vue+compiler-sfc@3.5.27_better-sqlite3@12.6.2_cac@6.7_d2aeb2292c1af4998397e345ba461360/node_modules/nuxt/dist/app/components/nuxt-stubs.js?v=9bbd892e' does not provide an export named 'default':undefined:undefined
```

mateusznarowski:
The clean reproduction I provided in the issue is now OK. I wanted to test the fix in my project where I use the `@nuxtjs/seo` module, and I get the following error

```
The requested module ‘/_nuxt/@fs/Users/mateusznarowski/temp/content-issue/node_modules/.cache/vite/client/deps/@unhead_schema-org_vue.js?v=67c7599a’ does not provide an export named 'SchemaOrgDebug'
```

To trigger the error, on a clean reproduction without any additional configuration simply add the module in `nuxt.config` and run dev.

mateusznarowski:
Before applying the patch, the component imports in the `./.nuxt/content/components.ts` folder look like this:

`export const CustomComponent = () => import(‘./../../app/components/content/CustomComponent.vue’)`

and after applying the patch, they look like this:

`export { default as CustomComponent } from ‘./../../app/components/content/CustomComponent.vue’`

I haven't analyzed this in depth, but won't this have an impact on performance? I also wondered what works differently in the `cloudflare_module` preset compared to others that such a change is required?

onmax:
Dynamic imports inside render functions don't work reliably in CF Workers runtime. Direct imports work, but lose lazy loading benefit.
There are other works around
- Keep lazy loading for dev, direct for CF Workers. Use preset detection to choose strategy:

```
const useLazyLoading = nuxt.options.nitro.preset !== 'cloudflare_module'
```
- Namespace imports in ContentRenderer. Keep lazy loading but use namespace import workaround for CF Workers:
```
 const comp = await import('./specific-component.vue')
```

I would go with the first one, but it is risky since is different behavior per env.

farnabaz:
Changing dynamic import into static import is not ideal solution, it will load all components when `ContentRenderer` is rendering. This is not ideal for large amount of components like Nuxt UI docs website.

We should find a way to sole dynamic import problem.

onmax:
Ready for review. I added two Cloudflare preview links in the PR body: main (fails SSR) and this branch (SSR fixed).

PR #3705, opened 2026-01-28
fix(queryCollectionSearchSections): fix options types

hendrikheil:
### 🔗 Linked issue
---

### ❓ Type of change

### 📚 Description

This PR updates the server and client types for `queryCollectionSearchSections`. It adds missing options like `minHeading`, `maxHeading` and `extraFields` and removes an unused options `separators`, which had no effect on the actual query.

I've also added a test for the extraFields behavior to ensure it actually works as expected.

It might be useful to export the options of `generateSearchSections` as a type so that they don't have to be duplicated two additional times?

### 📝 Checklist

### DIFF
--- a/src/runtime/client.ts
+++ b/src/runtime/client.ts
@@ -2,7 +2,7 @@ import type { H3Event } from 'h3'
 import { collectionQueryBuilder } from './internal/query'
 import { generateNavigationTree } from './internal/navigation'
 import { generateItemSurround } from './internal/surround'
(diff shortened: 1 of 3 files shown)


farnabaz:
> It might be useful to export the options of generateSearchSections as a type so that they don't have to be duplicated two additional times?


Thanks for the PR @hendrikheil 
Could you create this type and update PR? :)

hendrikheil:
@farnabaz Done, not sure if there is a convention for the names?

PR #3706, opened 2026-01-28
fix: align queryCollectionSearchSections opts type

onmax:
PR #3636 added `minHeading`/`maxHeading` to `generateSearchSections` but forgot to update the public API types.

Removes stale `separators` option, adds `minHeading`/`maxHeading` to match internal implementation.


hendrikheil:
I think our PRs are essentially the same ;-)

https://github.com/nuxt/content/pull/3705

onmax:
sorry didn't check!

Issue #3533, opened 2025-09-11
SQLite fails when UTF-8 multibyte character is split at exactly 350,000 bytes boundary


purupurupu:
### Environment

  - Operating System: Darwin
  - Node Version: v20.19.0
  - Nuxt Version: 3.16.2
  - Nuxt Content Version: 3.6.0
  - Nuxt Sitemap Version: 7.4.1
  - Better-SQLite3 Version: 11.10.0
  - Package Manager: yarn@4.2.2
  - Builder: -
  - User Config: -
  - Runtime Modules: -
  - Build Modules: -

### Version

v3

### Reproduction

1. Create a `sample.json` file in the `content` directory
2. Ensure that a UTF-8 multibyte character (e.g., Japanese character "心" = 0xE5 0xBF 0x83) starts at byte position 349,999

### json structure
```json
{
  "data": [
    {
      "id": "1",
      "content": "... [text padding to reach exactly 349,999 bytes] ... 中心 ..."
    }
  ]
}
```

### nuxt.config.ts
```ts
import { defineContentConfig, defineCollection, z } from '@nuxt/content'

export default defineContentConfig({
  collections: {
    content: defineCollection({
      type: 'data',
      source: '**/*.json',
      schema: z.object({}),
    }),
  },
})
```

### Description

When a UTF-8 multibyte character boundary falls exactly at byte position 350,000 in a JSON file, Nuxt Content's SQLite implementation fails with:

SQLite3Error: SQLITE_ERROR: unrecognized token: "[token_hash]"

And on the client side:
Unterminated string in JSON at position 349999 (line 1 column 350000)

### Investigation Results

Proof:
- Character "心" in UTF-8: 0xE5 0xBF 0x83 (3 bytes)
- At position 349,999: 0xE5 (first byte of "心")
- At position 350,000: 0xBF (second byte of "心") <- Buffer cuts here
- At position 350,001: 0x83 (third byte of "心")

### Verification

Tested with multiple files:
- File A: 685,647 bytes total, multibyte character at position 350,000 → ❌ Error
- File B: 3,150,308 bytes total, ASCII character at position 350,000 → ✅ Success
- File C: 3,751,769 bytes total, ASCII character at position 350,000 → ✅ Success

The error only occurs when position 350,000 falls within a multibyte UTF-8 character.

### Logs

```shell-script

```

farnabaz:
Thank you for reporting this issue. Could you please provide a minimal reproduction repository or online on Stackblitz?

[...]

PR #3717, opened 2026-02-12
fix(collection): use UTF-8 byte length for SQL query size check

yodakaEngineer:
### 🔗 Linked issue
Resolves https://github.com/nuxt/content/issues/3533

### ❓ Type of change

### 📚 Description

The SQL split logic in generateCollectionInsert used string.length  (character count) to check against Cloudflare D1's 100KB byte limit.
Multibyte characters (e.g. Japanese: 1 char = 3 bytes) could exceed the limit without triggering a split, causing SQLite errors on D1.
So, I replaced string.length with utf8ByteLength() for size comparisons and slice index calculations.

FYI: `心` means `heart`.

### 📝 Checklist

### DIFF
--- a/test/unit/generateCollectionInsert.test.ts
+++ b/test/unit/generateCollectionInsert.test.ts
@@ -167,4 +167,53 @@ describe('generateCollectionInsert', () => {
     }
     expect(reconstructed).toBe(content)
   })
+
+  test('Succeed when SLICE_SIZE byte boundary falls on an emoji', () => {
+    const collection = resolveCollection('content', defineCollection({
+      type: 'data',
+      source: '**',
+      schema: z.object({
+        content: z.string(),
+      }),
+    }))!
+
+    // 'a' (1 byte) shifts alignment so that the SLICE_SIZE byte boundary
+    // falls in the middle of a '😀' (4 bytes in UTF-8)
+    // biggestColumn = "'a😀😀...😀'" → byte 0: quote(1), byte 1: 'a'(1), bytes 2+: emojis(4 each)
+    // Byte at SLICE_SIZE (70000) = 2 + 4*17499.5 → falls inside the 17500th emoji
+    const content = 'a' + '😀'.repeat(25000)
+
+    const { queries: sql } = generateCollectionInsert(collection, {
+      id: 'emoji-boundary.md',
+      stem: 'emoji-boundary',
+      extension: 'md',
+      meta: {},
+      content,
+    })
+
+    // Must be split into multiple queries
+    expect(sql.length).toBeGreaterThan(1)
+
+    // Each query must fit within the byte limit
+    for (const query of sql) {
+      expect(utf8ByteLength(query)).toBeLessThan(MAX_SQL_QUERY_SIZE)
+    }
+
+    // First query should be INSERT, subsequent should be UPDATE
+    expect(sql[0]).toContain('INSERT INTO')
+    for (let i = 1; i < sql.length; i++) {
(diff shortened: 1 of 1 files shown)


yodakaEngineer:
I'm sorry.
Maybe, This will likely stabilize deployment to D1, but this PR cannot resolve https://github.com/nuxt/content/issues/3533.
I think it just looked like it fixed because the slice position changed.

yodakaEngineer:
Finaly, I got it!
Using .slice() to strings include emoji breaks character like below.

```js
'aa🎲aa'.slice(3) // This is '\uDFB2aa'
```

The current implementation breaks when an emoji appears at the `SLICE_SIZE` th character position.
I think this is cause of #3533.

This PR is changing how to size from character to byte, so I think this resolves #3533.
I also added test.
Please review it.

yodakaEngineer:
> Finaly, I got it! Using .slice() to strings include emoji breaks character like below.
> 
> ```js
> 'aa🎲aa'.slice(3) // This is '\uDFB2aa'
> ```
> 
> The current implementation breaks when an emoji appears at the `SLICE_SIZE` th character position. I think this is cause of #3533.
> 
> This PR is changing how to size from character to byte, so I think this resolves #3533. I also added test. Please review it.

FYI: This test is broken on main branch.

```ts
  test('Emoji at SLICE_SIZE boundary is not broken', () => {
    const collection = resolveCollection('content', defineCollection({
      type: 'data',
      source: '**',
      schema: z.object({
        content: z.string(),
      }),
    }))!

    // biggestColumn = `'${content}'` (with surrounding SQL quotes)
    // getSliceIndex checks column[SLICE_SIZE - 1] = column[69999]
    // column[0] = opening quote, so column[69999] = content[69998]
    // Place an emoji (surrogate pair, 2 UTF-16 code units) at content[69998]
    const prefix = 'a'.repeat(69998)
    const emoji = '😀'
    const suffix = 'b'.repeat(40000)
    const content = prefix + emoji + suffix

    const { queries: sql } = generateCollectionInsert(collection, {
      id: 'foo.md',
      stem: 'foo',
      extension: 'md',
      meta: {},
      content,
    })

    // Should be split into multiple queries
    expect(sql.length).toBeGreaterThan(1)

    // Each query should not contain broken surrogate pairs
    for (const query of sql) {
      console.log(query)
      for (let i = 0; i < query.length; i++) {
        const code = query.charCodeAt(i)
        if (code >= 0xD800 && code <= 0xDBFF) {
          // High surrogate must be followed by low surrogate
          const next = query.charCodeAt(i + 1)
          expect(next >= 0xDC00 && next <= 0xDFFF).toBe(true)
        }
        if (code >= 0xDC00 && code <= 0xDFFF) {
          // Low surrogate must be preceded by high surrogate
          const prev = query.charCodeAt(i - 1)
          expect(prev >= 0xD800 && prev <= 0xDBFF).toBe(true)
        }
      }
    }
  })
```

farnabaz:
Thanks for the PR @yodakaEngineer.

Don’t we need to reduce ‎`MAX_SQL_QUERY_SIZE`? If Japanese characters are 3 bytes, 100K characters might go above the 100KB Cloudflare D1 hard limit.

yodakaEngineer:
@farnabaz Thanks for quick response.

I think MAX_SQL_QUERY_SIZE is not character limit but byte limit.

But the existing implementation calculates based on character count, so this PR changes it to calculate based on byte count.

So, we don't need change MAX_SQL_QUERY_SIZE.

PR #3771, opened 2026-04-17
fix(preview): skip collections without source in preview template

### LINKED ISSUE #3767: previewTemplate serializes internal `info` collection's `source: undefined` as `source: []`
### Environment

- OS: macOS 15.4 (Darwin 25.4.0 arm64)
- Node: v24.13.0
- Nuxt: 4.4.2
- @nuxt/content: 3.12.0
- nuxt-studio: 1.6.0

### Version

3.12.0

### Reproduction

The bug is in the compiled source at `dist/module.mjs` and can be verified by inspecting the `/__preview.json` endpoint output. Steps:

1. Create a Nuxt 4 project with `@nuxt/content` v3 and `nuxt-studio`
2. Define any content collections in `content.config.ts`
3. Deploy with SSR and navigate to `/__preview.json`
4. Observe that the `info` collection has `"source": []` instead of being excluded or having `source` omitted

### Description

The internal `info` collection created by `resolveCollections()` has `source: void 0` (line 2405 of `dist/module.mjs`):

```js
collections.info = {
  type: "data",
  source: void 0,
  schema: infoStandardSchema,
  // ...
};
```

The `previewTemplate()` function (line 1701) serializes this into `source: []`:

```js
source: collection.source?.filter((source) => source.repository ? void 0 : collection.source) || [],
```

Since `undefined?.filter(...)` evaluates to `undefined`, the `|| []` fallback kicks in, producing `source: []` in the serialized output.

This causes downstream crashes in `nuxt-studio`, which iterates all collections and accesses `source[0].include` / `source[0].prefix` without guarding against empty arrays — resulting in `TypeError: Cannot read properties of undefined (reading 'include')`.

See also: https://github.com/nuxt-content/nuxt-studio/issues/425

#### Suggested fix

Either exclude collections without source from the preview payload:

```js
// In previewTemplate, skip the info collection
if (!collection.source) return acc;
```

Or use `??` instead of `||` to preserve the distinction:

```js
source: collection.source?.filter((source) => source.repository ? void 0 : collection.source) ?? undefined,
```

(With `?? undefined`, `JSON.stringify` will omit the key entirely.)

I'm happy to submit a PR if you'd prefer one approach over the other.

### Additional context

The `nuxt-studio` side also lacks null guards (tracked in https://github.com/nuxt-content/nuxt-studio/issues/425), but the root cause is the serialization here — `source: void 0` should not become `source: []` in the preview payload.

### Logs

```shell
TypeError: Cannot read properties of undefined (reading 'include')
    at parseSourceBase (source.js)
    at Array.map
    at async Object.list
    at async Object.load (main-CUjoLU7k.js)
```

humanbird:
### 🔗 Linked issue

Resolves #3767 (and unblocks nuxt-content/nuxt-studio#425)

### ❓ Type of change

### 📚 Description

The internal `info` collection (created by `resolveCollections()` in `src/utils/collection.ts`) has `source: undefined`. When `previewTemplate` serialized this collection it produced `source: []` in the preview manifest:

```ts
// before
source: collection.source?.filter(source => source.repository ? undefined : collection.source) || [],
```

`undefined?.filter(...)` → `undefined`, then the `|| []` fallback kicks in, so the info collection ends up with an empty `source` array — which violates the `ResolvedCollectionSource[]` contract (at least one source is expected) and crashes downstream consumers that read `source[0].include` and `source[0].prefix` (most visibly `nuxt-studio`, see nuxt-content/nuxt-studio#425).

This PR skips source-less collections entirely in the preview manifest, since they cannot be edited from the preview/studio UI anyway. It also simplifies the remote-collection filter to a straightforward `!source.repository` predicate (the previous `? undefined : collection.source` pattern was hard to read and the branch that returned `collection.source` was only truthy by accident).

### Before / After

For a Nuxt 4 project with `@nuxt/content` v3 plus any user-defined collection, `/__preview.json` previously contained:

```json
{
  "info": { "source": [], "type": "data", ... },
  "pages": { "source": [{ "include": "**", "prefix": "/", ... }], ... }
}
```

After this PR the `info` key is omitted entirely. User-visible collections are unaffected.

### 📝 Checklist

- [ ] I have updated the documentation accordingly. *(no docs change needed — internal behavior only)*

### DIFF
--- a/src/utils/templates.ts
+++ b/src/utils/templates.ts
@@ -206,13 +206,7 @@ export const previewTemplate = (collections: ResolvedCollection[], gitInfo: GitI
   filename: moduleTemplates.preview,
   getContents: ({ options }: { options: { collections: ResolvedCollection[] } }) => {
     const collectionsMeta = options.collections.reduce((acc, collection) => {
-      // Skip collections that would be serialized with an empty `source` array,
-      // because downstream consumers (e.g. nuxt-studio) iterate `source[0]` and
-      // crash on empty arrays. This covers two cases:
-      //   1. Collections with no source at all (e.g. the internal `info` collection).
-      //   2. Collections whose sources are all remote repositories — those are
-      //      filtered out below and should not appear in the preview manifest,
-      //      since they cannot be edited from the preview/studio UI either.
+      // Only include non remote collections and collections with at least one local source (remove `info` collection)
       const localSources = collection.source?.filter(source => !source.repository) ?? []
       if (localSources.length === 0) {
         return acc
--- a/src/utils/templates.ts
+++ b/src/utils/templates.ts
@@ -206,6 +206,14 @@ export const previewTemplate = (collections: ResolvedCollection[], gitInfo: GitI
   filename: moduleTemplates.preview,
   getContents: ({ options }: { options: { collections: ResolvedCollection[] } }) => {
     const collectionsMeta = options.collections.reduce((acc, collection) => {
+      // Skip collections without a source (e.g. the internal `info` collection).
+      // They cannot be edited from the preview/studio UI, and serializing them
+      // with an empty `source: []` array breaks downstream consumers that expect
+      // at least one `ResolvedCollectionSource` entry.
+      if (!collection.source) {
+        return acc
+      }
+
       const schemaWithCollectionName = {
         ...collection.extendedSchema,
         definitions: {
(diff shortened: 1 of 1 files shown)


humanbird:
Good catch by @coderabbitai — pushed [cba6554](https://github.com/nuxt/content/pull/3771/commits/cba6554c) to also cover the remote-only case. The logic now computes `localSources = collection.source?.filter(s => !s.repository) ?? []` up-front and skips the collection entirely when that's empty, so the invariant "a collection in the preview manifest has ≥ 1 local source" holds for both:

1. No source at all (internal `info` collection).
2. Sources exist but are all remote repositories.

larbish (approved):
✅

farnabaz (approved):
👍

PR #3749, opened 2026-03-21
fix(preview): move formatDate/formatDateTime into runtime subtree

daresTheDevil:
### 🔗 Linked issue

Resolves #3742

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

`src/runtime/internal/preview/collection.ts` imported `formatDate`/`formatDateTime` from `../../../utils/content/transformers/utils`. `nuxt-module-build` bundles `src/utils/` into the main chunk but emits `src/runtime/` files as separate modules — so that cross-tree import resolves to a path that doesn't exist in the published package.

Cloudflare Workers (strict module resolution) throws on the dangling import, breaking all v3.12 deploys on CF.

**Fix**: duplicate the two small functions into `src/runtime/internal/preview/utils.ts` and update the import in `collection.ts`. Added unit tests with a drift guard that asserts both copies produce identical output.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #3782, opened 2026-05-06
fix: use UTC getters in formatDate/formatDateTime to prevent timezone drift

surohak:
## Description

`formatDateTime` and `formatDate` use `new Date(input)` followed by local-time getters (`getFullYear`, `getMonth`, `getDate`, `getHours`, etc.). On CI runners or servers not running in UTC, this produces shifted dates.

For example, `formatDate("2023-01-01T00:30:00Z")` on a UTC-5 machine returns `"2022-12-31"` because 00:30 UTC is still Dec 31 in UTC-5.

Since these functions serialize values for D1/SQLite storage where dates should be treated as UTC, the local-time getters produce incorrect results.

This was also flagged in the code review of #3698 but was not addressed in the merge.

## Changes

Both copies of these functions are updated:
- `src/utils/content/transformers/utils.ts` (build-time)
- `src/runtime/internal/preview/utils.ts` (runtime, per #3749)

The fix:
1. Early-returns when input is already in canonical format (`YYYY-MM-DD` or `YYYY-MM-DD HH:mm:ss`) — avoids needless round-trip through `Date`.
2. Normalizes space-separated datetime strings to ISO format with `Z` suffix before parsing, ensuring UTC interpretation.
3. Replaces all local-time getters with UTC equivalents (`getUTCFullYear`, `getUTCMonth`, `getUTCDate`, `getUTCHours`, `getUTCMinutes`, `getUTCSeconds`).

## Test Plan

Updated `test/unit/formatDate.test.ts` with deterministic UTC assertions that pass regardless of the system timezone.


farnabaz (approved):
Thanks

PR #3783, opened 2026-05-06
fix: preserve null for optional booleans and strip empty JSON objects in refineContentFields

surohak:
## Description

Two issues with `refineContentFields` when processing D1/SQLite query results:

### 1. Optional boolean fields get incorrectly set to `false`

`Boolean(item[key])` is called unconditionally for boolean-typed fields. When a field is absent in the source document, SQLite stores NULL. After query: NULL → JS `null` → `Boolean(null) === false`.

This means truly-absent optional boolean fields get injected as `false` in the returned document, causing:
- Spurious diffs when comparing with the original YAML/markdown source
- Incorrect semantics (absence ≠ false for optional fields)

### 2. Empty JSON objects pollute the document

Optional JSON columns (e.g. `seo`) default to `'{}'` in D1. After parsing, this produces an empty `{}` on the document. The YAML source has no such key — this causes false change detection in Studio and inflates the document.

## Changes

`src/runtime/internal/collection.ts`:
- For booleans: if value is null/undefined, `delete item[key]` instead of coercing to false
- For JSON: after parsing, if the result is an empty object (and key is not `meta`), delete it

The `meta` field is exempted because `applyCollectionSchema` uses it as a catch-all bucket for unrecognized fields.

## Test Plan

- Documents with absent optional boolean fields should no longer show `field: false`
- Documents with empty JSON columns should not have `seo: {}` or similar empty objects
- The `meta` field continues to work as a catch-all (not stripped even when empty)

### DIFF
--- a/src/runtime/internal/collection.ts
+++ b/src/runtime/internal/collection.ts
@@ -1,6 +1,11 @@
 import type { CollectionInfo } from '@nuxt/content'
 import contentManifest from '#content/manifest'
 
+/**
+ * Refine raw fields from D1/SQLite query results into their proper JS types.
+ * Handles JSON parsing, boolean coercion (preserving null for absent fields),
+ * and removes empty JSON objects that arise from D1 column defaults.
+ */
 export function refineContentFields<T>(sql: string, doc: T) {
   const fields = findCollectionFields(sql)
   const item = { ...doc } as T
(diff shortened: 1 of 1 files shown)


farnabaz (changes requested):
Thanks for the PR @surohak,
Do you mind adding tests to verify the changes?

farnabaz:
Sorry for the late response @surohak, Could you add some tests to reproduce and cover these changes?

PR #3816, opened 2026-07-15
fix(dev): preserve SQL hashes in HMR dumps

danlg:
## Summary

- keep generated row hashes when HMR updates the compressed SQL dump
- keep browser HMR queries unchanged
- cover both matching collections in the existing HMR test

## Problem

The runtime importer expects every dump statement to end with ` -- <hash>`. HMR currently replaces the dump row with the raw INSERT, so the importer treats the whole INSERT as the hash, strips it, and can pass an empty SQL statement to SQLite (`The supplied SQL string contains no statements`).

## Test

`pnpm exec vitest run test/multi-collection-hmr.test.ts`


farnabaz (approved):
Thanks @danlg 
LGTM 👍

PR #3822, opened 2026-07-30
fix(nuxthub): handle object form of `hub.db`

### LINKED ISSUE #3821: nuxthub preset ignores the object form of hub.db and falls back to better-sqlite3
### Environment

- `@nuxt/content` 3.15.2
- `@nuxthub/core` 0.10.8
- Deployed on Vercel

### Description

The `nuxthub` preset only maps the NuxtHub database config into `content.database` when `hub.db` is a string. Both `setup()` and `setupNitro()` gate on:

```js
} else if (typeof nuxtOptions.hub?.db === "string" && typeof hubDb === "object") {
```

With `@nuxthub/core` >= 0.10 the config can be an object, for example:

```ts
hub: {
  db: {
    dialect: 'sqlite',
    applyMigrationsDuringBuild: false
  }
}
```

In that case the check fails, `content.database` silently falls back to the default `{ type: 'sqlite' }` and `resolveDatabaseAdapter` picks `db0/connectors/better-sqlite3` for the runtime adapter instead of libsql. The `file:/tmp/` rewrite in `setupNitro` is skipped too since the database type is no longer `libsql`.

On Vercel this makes every `POST /__nuxt_content/<collection>/query` return a 500:

```
Error: Module did not self-register: '/var/task/node_modules/better-sqlite3/build/Release/better_sqlite3.node'
    at bindings (/var/task/node_modules/bindings/bindings.js:112:48)
  code: 'ERR_DLOPEN_FAILED'
```

The addon builds and loads fine on the build image so prerendering works, it only fails in the lambda runtime. Rendered pages keep working, which makes the breakage invisible until something queries the endpoint server side, in our case the nuxt.com MCP server. Same symptom as nuxt/ui#5929 back in January.

This broke nuxt.com when we switched from `hub: { db: 'sqlite' }` to the object form in nuxt/nuxt.com#2354. Workaround is pinning the adapter explicitly:

```ts
content: {
  database: {
    type: 'libsql',
    url: 'file:/tmp/sqlite.db'
  }
}
```

The preset should probably normalize both forms, or read the resolved config from `runtimeConfig.hub.db` (which is always an object after `@nuxthub/core` resolves it) instead of gating on the raw user value.

benjamincanac:
### 🔗 Linked issue

Resolves #3821

### ❓ Type of change

### 📚 Description

The nuxthub preset only mapped the NuxtHub database config into `content.database` when `hub.db` was a string. With `@nuxthub/core` >= 0.10 the config can also be an object, in that case the check failed and `content.database` silently fell back to the default sqlite config, which picks `better-sqlite3` at runtime and breaks every query endpoint on Vercel with `ERR_DLOPEN_FAILED`.

Both `setup()` and `setupNitro()` now gate on `hub.db` being truthy and read the resolved config from `runtimeConfig.hub.db`, which is always an object with `driver` and `connection` once NuxtHub has run (both hooks fire after `modules:done`).

The duplicated driver mapping is extracted into `hubDatabaseToContentDatabase` and now covers the driver names NuxtHub 0.10 actually resolves to: `postgres-js` and `neon-http` map to `postgresql` (the old fallthrough produced an invalid `type: 'postgres-js'`), `better-sqlite3` maps to `sqlite`, and unsupported drivers (`mysql2`, `d1-http`) warn and fall back to the default database instead of producing an invalid config.

Added unit tests for the mapping, both `hub.db` forms, and the `/tmp` libsql rewrite in `setupNitro`.

### 📝 Checklist

### DIFF
--- a/src/presets/nuxthub.ts
+++ b/src/presets/nuxthub.ts
@@ -17,10 +17,11 @@ export function hubDatabaseToContentDatabase(hubDb: { driver: string, connection
     return { type: 'd1', bindingName: 'DB' }
   }
   if (['node-postgres', 'postgres-js', 'neon-http', 'postgres', 'postgresql'].includes(hubDb.driver)) {
-    return { type: 'postgresql', url: hubDb.connection?.url as string }
+    return hubDb.connection?.url ? { type: 'postgresql', url: hubDb.connection.url } : undefined
   }
   if (['sqlite', 'better-sqlite3'].includes(hubDb.driver)) {
-    return { type: 'sqlite', filename: (hubDb.connection?.filename as string) || (hubDb.connection?.url || '').replace(/^file:/, '') }
+    const filename = (hubDb.connection?.filename as string) || (hubDb.connection?.url || '').replace(/^file:/, '')
+    return filename ? { type: 'sqlite', filename } : undefined
   }
   if (['libsql', 'pglite'].includes(hubDb.driver)) {
     return { type: hubDb.driver, ...hubDb.connection } as unknown as ContentDatabaseConfig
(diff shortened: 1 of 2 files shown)


farnabaz (approved):
Thanks

PR #3826, opened 2026-08-06
fix(content): import zod-to-json-schema only for zod v3 schemas

### LINKED ISSUE #3825: Prefer zod v4's native toJSONSchema() — make the zod-to-json-schema import lazy so v4 consumers don't need it
### Context

Nuxt Content 3.15.2 already ships a native **zod v4** schema path: `dist/chunks/zod4.mjs` builds the JSON Schema via zod v4's built-in `z.toJSONSchema(schema, { target: "draft-7", ... })` (with a small date/`$content` override). The vendor is detected per schema (`dist/module.mjs:2334` — `schema.def ? "zod4" : "zod3"`).

### Problem

Despite the native v4 path, `dist/module.mjs:50` statically imports the v3-era converter at module top level:

```js
import { zodToJsonSchema, ignoreOverride } from "zod-to-json-schema";
```

and `dependencies` still pin the v3 line (`zod ^3.25.76`, `zod-to-json-schema ^3.25.2`).

Consequences for consumers that standardize on zod v4 (stable since 4.0):

- Because the import is static, `zod-to-json-schema` must be resolvable in the consuming app's tree for the module to load at all — even when every collection schema is zod v4 and the native `toJSONSchema()` path is the one actually used.
- Docus-based apps therefore must declare both `zod` and `zod-to-json-schema` purely to satisfy Nuxt Content's module load; `nuxt prepare` fails without them (a recurring trap — see the comment in geoql/v-maplibre's `pnpm-workspace.yaml`).
- Zod v4 itself makes `zod-to-json-schema` obsolete for this use case: `z.toJSONSchema()` is built in.

### Suggestion

1. Make the `zod-to-json-schema` import **lazy** — dynamic-import it inside the zod3 `toJSONSchema` path only (the zod4 path already uses the native API), so v4-only consumers never need the package in their tree.
2. Widen the zod dependency to include v4 (`>=3.25.0 <5`), and move zod/zod-to-json-schema to (optional) peerDependencies so the version the app actually uses is respected.

Happy to open a PR if this direction is welcome.

lazerg:
### 🔗 Linked issue

Closes #3825

### ❓ Type of change

### 📚 Description

`initiateValidatorsContext()` imports the zod3 validator dynamically, behind an `isPackageInstalled('zod')` check, but `src/utils/index.ts` re-exports `z` from that same module. The static re-export wins, so `dist/module.mjs` ends up with a top-level `import { zodToJsonSchema, ignoreOverride } from 'zod-to-json-schema'` and the module cannot load at all unless that package resolves, even for apps whose collections are all zod v4 and go through the native `toJSONSchema()` path.

The legacy `z` export and its `.editor()` prototype patch move to `schema/zod3-legacy.ts`, which leaves `schema/zod3.ts` as a converter-only module like `zod4.ts` and `valibot.ts`. It stays out of the eager graph and is loaded only when both `zod` and `zod-to-json-schema` are present, so a missing converter now surfaces as the existing "install `zod` and `zod-to-json-schema`" error instead of a module resolution crash.

This only covers the first suggestion in the issue. Widening the zod range and moving zod/zod-to-json-schema to optional peer dependencies changes the dependency contract, so I left it out; `zod` itself is still imported eagerly for the legacy `z` export.

### 📝 Checklist

### DIFF
--- a/src/utils/dependencies.ts
+++ b/src/utils/dependencies.ts
@@ -78,6 +78,8 @@ export async function initiateValidatorsContext() {
     nuxtContentContext().set('valibot', await import('./schema/valibot'))
   }
   if (await isPackageInstalled('zod')) {
+    // The zod3 converter statically imports `zod-to-json-schema`, so only
+    // register it when that package is actually reachable.
     if (await isPackageInstalled('zod-to-json-schema')) {
       nuxtContentContext().set('zod3', await import('./schema/zod3'))
     }
--- a/src/utils/dependencies.ts
+++ b/src/utils/dependencies.ts
@@ -78,7 +78,9 @@ export async function initiateValidatorsContext() {
     nuxtContentContext().set('valibot', await import('./schema/valibot'))
   }
   if (await isPackageInstalled('zod')) {
-    nuxtContentContext().set('zod3', await import('./schema/zod3'))
+    if (await isPackageInstalled('zod-to-json-schema')) {
+      nuxtContentContext().set('zod3', await import('./schema/zod3'))
+    }
     nuxtContentContext().set('zod4', await import('./schema/zod4'))
   }
 }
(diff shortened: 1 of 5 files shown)


lazerg:
Hi @farnabaz, just checking if you have had a chance to look at this yet. Happy to make any changes.

lazerg:
Closing this since it has been inactive for a while.

PR #3835, opened 2026-08-18
fix(ContentRenderer): keep async component identity stable across re-resolves

hendrikheil:
### 🔗 Linked issue

Ref nuxt-content/nuxt-studio#383

### ❓ Type of change

### 📚 Description

`defineAsyncComponent` returns a **new component type** on every call. `resolveVueComponent` calls it in two places, and it is reached from `componentsMap`, a `computed` that re-resolves whenever `props.value` changes identity:

```js
const componentsMap = computed(() => {
  return body.value ? resolveContentComponents(body.value, { tags: tags.value }) : {}
})
```

```js
else if (localComponents.includes(pascalCase(component))) {
  const loader = localComponentLoaders[pascalCase(component)]
  _component = loader ? defineAsyncComponent(loader) : undefined  // new type every time
}
```

So every re-resolve hands `MDCRenderer` a fresh component type for each local component. Vue sees a different type at the same vnode position, so it cannot patch — it unmounts and remounts, and the async wrapper renders nothing for a tick while it re-resolves.

This is what makes Studio's live preview flash the entire article on every keystroke: Studio calls `app:data:refresh`, `useAsyncData` hands back a new document object, and the whole rendered tree is destroyed and recreated. Because the DOM is torn down, page height momentarily collapses and the scroll position is lost too — that's the "jumps back to the top" half of nuxt-content/nuxt-studio#383.

Projects that mark their content components `global` don't hit this, because `resolveComponent` returns a stable definition. Local components are a supported (and bundle-size-motivated) configuration, so they should behave the same.

The fix caches both wrappers — a `Map` keyed by pascal name for the loader branch, a `WeakMap` keyed by the component object for the `setup` branch — so repeated resolutions return the same instance and Vue patches the tree.

I also hoisted the repeated `pascalCase(component)` into a local, since it's now used three times.

### 📝 Checklist

- [x] I have added tests (none — this is a component-identity fix with no observable output change; happy to add a regression test if you'd like one, though asserting "same vnode type across re-resolves" needs a bit of test scaffolding that doesn't exist yet).

### DIFF
--- a/src/runtime/components/ContentRenderer.vue
+++ b/src/runtime/components/ContentRenderer.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { kebabCase, pascalCase } from 'scule'
 import { resolveComponent, toRaw, defineAsyncComponent, computed } from 'vue'
+import type { Component } from 'vue'
 import type { MDCComment, MDCElement, MDCRoot, MDCText } from '@nuxtjs/mdc'
 import htmlTags from '@nuxtjs/mdc/runtime/parser/utils/html-tags-list'
 import MDCRenderer from '@nuxtjs/mdc/runtime/components/MDCRenderer.vue'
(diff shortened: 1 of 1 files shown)


nicokempe:
Hi @hendrikheil with that change, edits in Nuxt Studio don't seem to update the live preview at all anymore? No flicker and no scroll-issues anymore as described in https://github.com/nuxt-content/nuxt-studio/issues/383 but now no changes are visible in the front-end, updates are written to the file but a page reload doesn't even help to make them visible - does it work in your project? I can also share a reproduction video again if that's helpful

farnabaz (approved):
Thanks

nicokempe:
> Hi @hendrikheil with that change, edits in Nuxt Studio don't seem to update the live preview at all anymore? No flicker and no scroll-issues anymore as described in [nuxt-content/nuxt-studio#383](https://github.com/nuxt-content/nuxt-studio/issues/383) but now no changes are visible in the front-end, updates are written to the file but a page reload doesn't even help to make them visible - does it work in your project? I can also share a reproduction video again if that's helpful

@hendrikheil did you see this response? Did it work for you?

farnabaz:
@nicokempe Thanks for the ping! I tested it with my local playground example, and it was working as expected. Is your project open source, or are you able to reproduce the issue in a simple way? I can look into it if you can provide a reproduction. 🙂

nicokempe:
Tried it again yesterday and seems to work now (probably a caching issue on my side), thanks for the fix! 🙂

PR #3837, opened 2026-08-19
fix(module): keep content templates when building with `_prepare`

### LINKED ISSUE #3836: nuxt prepare / typecheck drops content tables of a running dev server
### Environment

- Nuxt 4.5.2, @nuxt/content 3.15.2
- Node 24.16.0, macOS 26.6.1
- Default SQLite connector (better-sqlite3)

### Version

@nuxt/content 3.15.2

### Reproduction

I have no hosted reproduction, but the steps are short and work on any project with a collection:

1. Start `nuxt dev` and open a page that queries a collection. It works.
2. In a second terminal run `npx nuxt prepare`. (`nuxt typecheck` runs the same build first.)
3. Reload the page.

The page returns 404 now. The table of the collection you queried is dropped and is not created again. Only a restart of the dev server brings the content back.

I can put together a minimal repo if that helps.

### Description

`nuxt prepare` and `nuxt typecheck` call `buildNuxt()` with `_prepare: true`, and they build into the same `.nuxt` folder the running dev server uses.

In `_prepare` mode the module returns before it registers the `modules:done` hook, so `processCollectionItems()` never runs and `manifest.dump` stays empty. The dump templates (`content/sql_dump.txt` and `content/database.compressed.mjs`) are still written, so they are written empty.

The dev server picks up the new templates. On the next query the integrity check sees a different checksum, drops the collection table and tries to import the dump. The dump is empty, `decompressSQLDump()` throws, and the table is never created again.

Only the collection that is queried first gets dropped, so at first it looks like one random collection disappeared.

A separate database for the build does not help here, because the problem is the shared `.nuxt` folder, not the SQLite file.

Would it make sense to skip the dump templates when `nuxt.options._prepare` is set, so the previous ones stay in place?

### Logs

```shell-script
ERROR Database integrity check failed The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined

    at Buffer.from (node:buffer:362:9)
    at decompressSQLDump (.nuxt/dev/index.mjs)
    at async _checkAndImportDatabaseIntegrity (.nuxt/dev/index.mjs)
    at async checkAndImportDatabaseIntegrity (.nuxt/dev/index.mjs)

ERROR [request error] [unhandled] [POST] /__nuxt_content/articlesRu/query
Error: no such table: _content_articlesRu
    at Object.all (node_modules/@nuxt/content/dist/runtime/internal/database.server.js:20:29)
    at Object.handler (node_modules/@nuxt/content/dist/runtime/api/query.post.js:13:35)
```

mansurkhoja:
### 🔗 Linked issue

Resolves #3836

### ❓ Type of change

### 📚 Description

`nuxt prepare` and `nuxt typecheck` build with `_prepare: true`. In that mode the module returns before it registers the `modules:done` hook, so collections are never indexed and `manifest.dump` and `manifest.checksum` stay empty. The templates are still written, so they are written from an empty manifest.

This is a problem when a dev server runs on the same build dir, which is the normal case when you type-check while working. The dev server picks up the new templates. On the next query the integrity check sees an empty checksum, drops the collection table and tries to import the dump. The dump is empty too, `decompressSQLDump()` throws, and the table is never created again. All content of that collection returns 404 until the dev server is restarted.

This PR keeps the previous file for the templates that are built from the manifest: both dumps, the per collection dump, the manifest itself and the components list. When there is no previous file (a clean build dir) nothing changes.

I have no test for it, because it needs a dev server and a second build next to it. I checked it by hand on the reproduction from #3836: with the patch, two `nuxt prepare` runs in a row leave the dev server and its content untouched, and `.nuxt/content/sql_dump.txt` keeps the real dump instead of becoming empty.

### 📝 Checklist

### DIFF
--- a/src/utils/templates.ts
+++ b/src/utils/templates.ts
@@ -1,5 +1,6 @@
+import { readFile } from 'node:fs/promises'
 import { gzip } from 'node:zlib'
-import type { NuxtTemplate } from '@nuxt/schema'
+import type { Nuxt, NuxtTemplate } from '@nuxt/schema'
 import { isAbsolute, join, relative } from 'pathe'
 import { compile as jsonSchemaToTypescript } from 'json-schema-to-typescript-lite'
 import type { JSONSchema } from 'json-schema-to-typescript-lite'
@@ -17,6 +18,20 @@ const compress = (text: string): Promise<string> => {
     }
     return resolve(buff?.toString('base64'))
   }))
+}
+
+/**
+ * Read what the last build wrote for this template.
+ *
+ * `nuxt prepare` and `nuxt typecheck` build with `_prepare: true`, and in that
+ * mode collections are not indexed: the manifest carries no dump and no
+ * checksums. Writing the templates from it would tell the app that there is no
+ * content at all. A dev server that shares the same build dir picks the files
+ * up, fails the integrity check and drops its collection tables, so its content
+ * is gone until it is restarted. Keeping the previous file avoids that.
+ */
+async function contentsOfPreviousBuild(nuxt: Nuxt, filename: string) {
+  return await readFile(join(nuxt.options.buildDir, filename), 'utf8').catch(() => undefined)
 }
 
 function indentLines(str: string, indent: number = 2) {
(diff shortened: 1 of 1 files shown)


farnabaz (approved):
LGTM 👍

PR #3838, opened 2026-08-19
fix(query): tolerate trailing slashes in path-based lookups

danielroe:
### 🔗 Linked issue

context: https://github.com/nuxt/nuxt/issues/36111
https://gitlab.com/eu-os/eu-os.gitlab.io/-/merge_requests/19

### ❓ Type of change

### 📚 Description

nuxt/content is sensitive to trailing slashes, which means that any user passing `route.path` on a deployment which _adds_ trailing slashes will lead to a 404 until nuxt redirects them to a version of the page without a trailing slash.

fwiw, `.path()` was already trailing-slash tolerant, so I think the fact that query wasn't is unintentional rather than a design decision

### 📝 Checklist


farnabaz (approved):
Thanks

